PR #8, opened 2015-01-22
use `is-md` module

sindresorhus:
because npm

https://github.com/sindresorhus/is-md


azu:
@sindresorhus Good! Thanks!

PR #35, opened 2015-10-11
docs(txtnode): why line of location start with 1?

azu:
Add docs about location indexed.

```
    interface Position {
        line: number; // start with 1
        column: number;// start with 0
        // This is for compatibility with JavaScript AST.
        // https://gist.github.com/azu/8866b2cb9b7a933e01fe
    }
```

- `line` - 1-indexed
- `column` - 0-indexed


azu:
This is **reverted** by #38 #40

Issue #37, opened 2015-10-11
Processor Plugin: Custome Parser


azu:
We want to support custom parser.

e.g.) textlint can lint `.rst` file by custom parser.

- [x] custom pre/post process(parse/modified) #36 
- [ ] custom AST traverse and Syntax => lator
- [x] customize support file extensions #36 

These are needed to realize.

azu:
## use case

- new format support
- middleware format
     - markdown -> ast -> cst -> rule

azu:
Please see https://github.com/textlint/textlint/blob/master/docs/plugin.md#processoroptional

PR #51, opened 2015-10-27
HTML support

azu:
#37 with HTML parse plugin

- experimental support


azu:
will move this to separated module.

PR #82, opened 2015-11-14
fix: use interop-require instead of require for dynamic loading

### LINKED ISSUE #81: Definition for rule \'' + key + '\' was not found. on Babel 6 generated rule
Babel 6 kill `module.exports = x` behavior.

- [Kill CommonJS default export behaviour · Issue #2212 · babel/babel](https://github.com/babel/babel/issues/2212 "Kill CommonJS default export behaviour · Issue #2212 · babel/babel")

It means that doesn't resolve dynamic `require(name)`.

textlint use [dynamic loading](https://github.com/textlint/textlint/blob/c4555df1083fb4ac3e61111cd4d74b1a0a99db2a/src/textlint-engine.js#L159) for rule and plugin.

We should use [interop-require](https://github.com/izaakschroeder/interop-require "interop-require") instead of `require` function.

azu:
close #81

### DIFF
--- a/test/fixtures/rules/issue81/no-default-assign-rule.js
+++ b/test/fixtures/rules/issue81/no-default-assign-rule.js
@@ -1,0 +1,12 @@
+// LICENSE : MIT
+"use strict";
+// This module has not module.export, but has module.export.default
+// See https://github.com/textlint/textlint/issues/81
+module.exports.__esModule = true;
+module.exports.default = function (context) {
+    const exports = {};
+    exports[context.Str] = function (node) {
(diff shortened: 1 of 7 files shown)


test/fixtures/rules/issue81/no-default-assign-rule.js
```diff
@@ -0,0 +1,12 @@
+// LICENSE : MIT
+"use strict";
+// This module has not module.export, but has module.export.default
+// See https://github.com/textlint/textlint/issues/81
+module.exports.__esModule = true;
```
azu: simulate Babel's output code.

Issue #91, opened 2015-11-19
Enhance document(ES6)


azu:
- [x] Write example as ES6
- [x] Add `RuleError` options
- [x] Add FAQ about "style" linting => recommend https://github.com/wooorm/mdast-lint 
- [x] Quick Start Guide
- [x] Usage of preset #96

PR #94, opened 2015-12-02
docs: ES6 based example

azu:
Rewrite all example with ES6.

- [x] remove `rules/` ?

Reflated #91


docs/configuring.md
```diff
@@ -79,10 +79,10 @@ it mean that
 
 Can use the following format:
 
-```js
+```json
 {
   "rules": {
-    "<rule-name>": true | false | object
+    "<rule-name>": true || false || object
```
azu: require more readable syntax

package.json
```diff
@@ -29,8 +29,10 @@
     "build": "babel src --out-dir lib --source-maps",
     "watch": "babel src --out-dir lib --watch --source-maps",
     "prepublish": "npm run --if-present build",
+    "lint:doc": "eslint docs/ --ext md",
+    "lint:js": "jscs src",
     "test:example": "./test/example-test.sh",
-    "test": "mocha test/*.js && jscs src && npm run test:example",
+    "test": "mocha test/*.js && npm run lint:doc && npm run lint:js && npm run test:example",
```
azu: use npm run-all?

PR #127, opened 2016-02-27
add --init option like eslint.

takahashim:
with `--init` option, generate .textlintrc file if not existed.

### DIFF
--- a/bin/textlint.js
+++ b/bin/textlint.js
@@ -4,8 +4,7 @@ var concat = require("concat-stream"),
     cli = require("../lib/cli");
 
 var exitCode = 0,
-    useStdIn = (process.argv.indexOf("--stdin") > -1),
-    init = (process.argv.indexOf("--init") > -1);
+    useStdIn = (process.argv.indexOf("--stdin") > -1);
 
 if (useStdIn) {
     process.stdin.pipe(concat({encoding: "string"}, function (text) {
@@ -14,12 +13,6 @@ if (useStdIn) {
             process.exit(exitCode);
         });
     }));
-} else if (init) {
-    var configInit = require("../lib/config/config-initializer");
-    configInit.initializeConfig().then(function (code) {
-        exitCode = code;
-        process.exit(exitCode);
-    });
 } else {
     cli.execute(process.argv).then(function (code) {
         exitCode = code;
--- a/bin/textlint.js
+++ b/bin/textlint.js
@@ -4,7 +4,8 @@ var concat = require("concat-stream"),
     cli = require("../lib/cli");
 
 var exitCode = 0,
-    useStdIn = (process.argv.indexOf("--stdin") > -1);
+    useStdIn = (process.argv.indexOf("--stdin") > -1),
+    init = (process.argv.indexOf("--init") > -1);
 
 if (useStdIn) {
     process.stdin.pipe(concat({encoding: "string"}, function (text) {
@@ -13,6 +14,12 @@ if (useStdIn) {
             process.exit(exitCode);
         });
     }));
+} else if (init) {
+    var configInit = require("../lib/config/config-initializer");
+    configInit.initializeConfig().then(function (code) {
+        exitCode = code;
+        process.exit(exitCode);
+    });
 } else {
     cli.execute(process.argv).then(function (code) {
         exitCode = code;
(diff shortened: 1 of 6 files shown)


bin/textlint.js
```diff
@@ -4,7 +4,8 @@ var concat = require("concat-stream"),
     cli = require("../lib/cli");
 
 var exitCode = 0,
-    useStdIn = (process.argv.indexOf("--stdin") > -1);
+    useStdIn = (process.argv.indexOf("--stdin") > -1),
+    init = (process.argv.indexOf("--init") > -1);
```
azu: Remove this from `bin/`
Instead, Add `init` option to [options.js](https://github.com/textlint/textlint/blob/master/src/options.js "options.js") like [this](https://github.com/eslint/eslint/blob/8ca6bedac12155891d5dca556498ce12a035eccb/lib/options.js#L177-L182).

bin/textlint.js
```diff
@@ -13,6 +14,12 @@ if (useStdIn) {
             process.exit(exitCode);
         });
     }));
+} else if (init) {
+    var configInit = require("../lib/config/config-initializer");
+    configInit.initializeConfig().then(function (code) {
```
azu: It should be written in [src/cli.js](https://github.com/textlint/textlint/blob/master/src/cli.js "src/cli.js").

```js
    execute(args, text) {
        var currentOptions;
        try {
            currentOptions = options.parse(args);
        } catch (error) {
            console.error(error.message);
            return Promise.resolve(1);
        }
        const files = currentOptions._;
        if (currentOptions.version) {
            // version from package.json
            console.log(`v${ require('../package.json').version }`);
        } else if (currentOptions.init) {
            // TODO: --init handling 
            return configInit.initializeConfig(process.cwd());
        } else if (currentOptions.help || !files.length && !text) {
            console.log(options.generateHelp());
        } else {
            // specify file name of stdin content
            const stdinFilename = currentOptions.stdinFilename;
            debug(`Running on ${ text ? 'text' : 'files' }, stdin-filename: ${stdinFilename}`);
            return this.executeWithOptions(currentOptions, files, text, stdinFilename);
        }
        return Promise.resolve(0);
    },
```

src/config/config-initializer.js
```diff
@@ -0,0 +1,33 @@
...
+const path = require('path');
+const Config = require('../config/config');
+
+// Public Interface
+/**
+ * Initializer class for config of textlint.
+ */
+const init = {
+    /**
+     * Create .textlintrc file in the current working directory
```
azu: I think that *current working directory* should be optional.
It means that `initializeConfig(dir)` accept `dir` option.
It is useful for testing.

src/config/config-initializer.js
```diff
@@ -0,0 +1,33 @@
...
+     * @returns {int} The exit code for the operation.
+     */
+    initializeConfig() {
+        let rcFile = "." + Config.CONFIG_FILE_NAME + "rc";
+        let output = JSON.stringify({"rules":{}}, null, '  ');
```
azu: [JSON.stringify(value[, replacer[, space]])](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify "JSON.stringify(value[, replacer[, space]])")'s `space` options is String or Number.
I prefer Number. `JSON.stringify({"rules":{}}, null, 2);`

takahashim:
Thanks for your comments in detail!
I'll fix and rebase them soon.

test/config-initializer-test.js
```diff
@@ -0,0 +1,61 @@
+// LICENSE : MIT
+'use strict';
+var assert = require("power-assert");
+var Config = require("../src/config/config");
+var configInit = require("../src/config/config-initializer");
+var loadConfig = require("../src/config/config-loader");
+var path = require("path");
+var os = require("os");
+var sh = require("shelljs");
```
azu: `shelljs` is missing in package.json.
Please do `npm i -D shelljs`

takahashim:
sorry, I'm installing v0.12 now...

azu:
@takahashim `npm i -D shelljs` resolve this error, maybe.
(Node v5 use npm@3 which apply flat dependencies.)

takahashim:
Thanks, I did so.

azu:
@takahashim Thanks for work :+1: 
I'v updated document and release [5.4.0](https://github.com/textlint/textlint/releases/tag/5.4.0 "5.4.0").

takahashim:
Thanks to merge!

PR #135, opened 2016-03-06
feat(RuleError): Index-based report

### LINKED ISSUE #134: context.report(node, { index }): index-based report
We want to support `context.report(node, { index: index }):`.

Currently, textlint's `RuleError` support `line` and `column`.

- [textlint/rule.md at master · textlint/textlint](https://github.com/textlint/textlint/blob/master/docs/rule.md "textlint/rule.md at master · textlint/textlint")

e.g.) It is wrong that `column` doen't consider line break.

```js
//  surrogate pair 
function stringToArray(value) {
    return value.match(/[\uD800-\uDBFF][\uDC00-\uDFFF]|[^\uD800-\uDFFF]/g) || [];
}

export default function (context) {
    let {Syntax, RuleError, report, getSource} = context;
    return {
        [Syntax.Str](node){
            const text = getSource(node);
            const strArray = stringToArray(text);
            for (let index = 0; index < strArray.length; index++) {
                const item = strArray[index];
                if (/❌/.test(item)) {
                    report(node, new RuleError("Use X insteadof ❌", {
                          column: index
                    }));
                }
            }
        }
    }
}
```

Input: 

```
string
❌
string
```

Result of TextLintMessage:

```json
{
   "message": "Use X insteadof ❌",
   "line": 1,
   "column": 7
}
```

Expected:

```json
{
   "message": "Use X insteadof ❌",
   "line": 2,
   "column": 1
}
```

We want to write this like

```js
//  surrogate pair 
function stringToArray(value) {
    return value.match(/[\uD800-\uDBFF][\uDC00-\uDFFF]|[^\uD800-\uDFFF]/g) || [];
}

export default function (context) {
    let {Syntax, RuleError, report, getSource} = context;
    return {
        [Syntax.Str](node){
            const text = getSource(node);
            const strArray = stringToArray(text);
            for (let index = 0; index < strArray.length; index++) {
                const item = strArray[index];
                if (/❌/.test(item)) {
                    report(node, new RuleError("Use X insteadof ❌", {
                          index: index // <= index
                    }));
                }
            }
        }
    }
}
```

azu:
Fix #134


src/rule/source-code.js
```diff
@@ -27,6 +28,10 @@ export default class SourceCode {
         assert(ext || filePath, "should be set either of fileExt or filePath.");
         this.hasBOM = text.charCodeAt(0) === 0xFEFF;
         this.text = (this.hasBOM ? text.slice(1) : text);
+        /**
+         * @type StructuredSource
+         */
```
azu: StructuredSource can convert loc to/from range

src/rule/source-location.js
```diff
@@ -0,0 +1,123 @@
...
+    adjust(node, padding) {
+        const nodeRange = node.range;
+        let line = node.loc.start.line;
+        let column = node.loc.start.column;
+        /*
+            FIXME: It is old way and un-document way
+            new RuleError("message", index);
+         */
+        let _paddingIndex;
+        if (typeof padding === "number") {
```
azu: This is un-document. Should be removed next version.

src/rule/source-code.js
azu: We'll create migration tools.

Issue #110, opened 2016-01-31
Contributing Guide


azu:
We should add `CONTRIBUTING.md`.

Ref:

- [Contributing Guidelines](https://github.com/blog/1184-contributing-guidelines "Contributing Guidelines")
- http://eslint.org/docs/developer-guide/contributing/working-on-issues

azu:
- [ ] Add link from Getting Started

takahashim:
@azu Does this project have coding standard?
I'm not sure about eslint so much, but if this project has `.eslintrc` or something, can we use it to check against convention?

azu:
@takahashim We have [.jscsrc](https://github.com/textlint/textlint/blob/master/.jscsrc ".jscsrc") file.
[JSCS](https://github.com/jscs-dev/node-jscs "JSCS")(JavaScript Code Style checker) check codeing style.

You can run JSCS via

```
npm run lint:js
# or
npm test
# automatically fix issue
$(npm bin)/jscs -x src
```

takahashim:
@azu Thanks!  I didn't know it.

(Note: I don't know JavaScript well, especially ES5 and ES6)

What kind of conventions I want to know are, for example:

* Which should I use for quotation of string, such as `require("foo")` or `require('foo')`  (I think both are used, and `"validateQuoteMarks"` is null in `.jscsrc`)
* Which (and when) should I use `var` or `let` (I think both are used)
* How to write comments; which (and when) should I use `//` and `/* ... */`

Can JSCS check these things?   If not, it's very helpful for me to refer to such things in contributing guide.

azu:
> Which should I use for quotation of string

prefer `"`(double quote)

> "validateQuoteMarks" is null in .jscsrc

Yes, it's temporary disabled...

> Which (and when) should I use var or let

`const` > `let` >>> `var`

> How to write comments

I'v used both style. I think that is no problem.

> Can JSCS check these things? 

Probably, ESLint can check these things.
We consider moving to ESLint :thought_balloon: 

> If not, it's very helpful for me to refer to such things in contributing guide.

Ya. I will take it after implementing fixer #131

azu:
I'v create pull request to #139

azu:
We have added [Contributing Guide](https://github.com/textlint/textlint/blob/master/CONTRIBUTING.md "Contributing Guide")!

PR #137, opened 2016-03-07
style(eslint): use ESLint insteadof JSCS

azu:
style(eslint): use ESLint insteadof JSCS

ref #110


azu:
- [ ] `"prefer-const": "error"`

src/rule/source-location.js
```diff
@@ -72,52 +75,78 @@ report(node, new RuleError("message", {
...
+        {
+            line,
+            column
+            fix?
+        }
+         */
+        return ObjectAssign({}, adjustedLoc, adjustedFix);
+    }
+
+    _adjustLoc(node, padding, _paddingIndex) {
```
azu: This has regission.
I'll fix it

azu: fixed in https://github.com/textlint/textlint/commit/6550bf5bdbeba4d13d1e5b802011badff43cc9e2

PR #140, opened 2016-03-09
feat(config): --no-color support

### LINKED ISSUE #105: --no-color option does not work when using pretty-error format
I just noticed. Though I have no trouble with this issue, output of `textlint --no-color -f pretty-error` is colorized as the following screenshot:

Note that `--no-color` option works as expected in stylish format.

azu:
Fix #105 by this and https://github.com/textlint/textlint-formatter/pull/8


azu:
Do not add feature and refactor at once

Issue #149, opened 2016-03-13
--fix and dry-run


azu:
## Issue

When we use `--fix`,

- Which is default behavitor, overwrite or dry-run?

azu:
Use case 1

1. lint some file
    - show "how to fix this"
2. --fix some file
    - if the user want to run with dry. use `--dry-run`

azu:
I added `--dry-run` option.
Closed by #150

PR #150, opened 2016-03-13
--formatter diff & --dry-run

azu:
Add a formatter and `--dry-run` for fixer.

- `--formatter diff` 
- `--dry-run` #149 

fix-dry-run zsh 2016-03-13 23-03-38

I think that `--formatter diff` to be default


azu:
@homu r+

PR #152, opened 2016-03-14
docs(FAQ): failed to load modules

azu:
> Failed to load textlint's rule module: "${packageName}" is not found.

See FAQ: https://github.com/textlint/textlint/blob/master/docs/faq/failed-to-load-textlints-module.md


azu:
@homu r+

FAQ document is inspired by deku v1.

PR #156, opened 2016-03-15
chore(readme): add chrome extension

azu:
Add URL of @io-monad's chrome extension.


azu:
@homu r+ 
We want to merge.

Issue #141, opened 2016-03-10
 Compliances tests for TxtNode


azu:
`TxtNode` is defined as docs/

- https://github.com/textlint/textlint/blob/master/docs/txtnode.md

We'll create `textlint-txt-node-tester` module.
It is similar to Promise A/+ tester and adapter approach.

- [promises-aplus/promises-tests: Compliances tests for Promises/A+](https://github.com/promises-aplus/promises-tests)
- [zenparsing/es-observable: Observables for ECMAScript](https://github.com/zenparsing/es-observable)

The motivation of the module is that plugin creator easy to create new plugin.

Any thought?

Related #122 

### Next Step

- Rename [TxtSyntax](https://github.com/textlint/textlint/blob/master/docs/txtnode.d.ts "TxtSyntax") to `TxtAST`
- Define `value` to `TxtNode`
     - Split `TxtNode` to `ParentNode` and `InlineNode`
- Add note about https://github.com/wooorm/unist

azu:
https://github.com/textlint/textlint/blob/master/docs/txtnode.md#txtnode-interface-1 don't define `value` property.
I think `value` is depended on the Syntax, but it is useful for some utility.

- https://github.com/azu/textlint-util-to-string

We want to define **What is value?**

By contrast, 

Q. What is `raw`?
A. `raw` is the vaue of Node's `range`.

azu:
My answer to "What is value?"

A. Inline Node should have `value`. It is plain vaue removed decoration.

azu:
I thinks that DOM Node's specification have the answer to "What is value?".

azu:
Idealistically, it is good that will be compatible for https://github.com/wooorm/unist

azu:
I notice that current TxtNode is compatible for unist?

Difference:

- `TxtNode` is not defiend [Text](https://github.com/wooorm/unist#text "Text"), but we will define this in the issue.
- `location` on Unist is `loc` in TxtNode.
- [Position](https://github.com/wooorm/unist#position "Position")'s column is 1-indexed, but TxtNode's positon is 0-indexed.

wooorm:
As an aside, why not use Unist directly? A benefit would be that you could use all the already existing  utilities.

azu:
> As an aside, why not use Unist directly?

I think that backword-compatibility for textlint's rule is important.

wooorm:
Hmm, okay. I’d still suggest to do so in a major release though :)

azu:
[...]

PR #170, opened 2016-03-19
docs(txtnode): Update document about TxtNode/TxtAST

azu:
Reflect update from "Compliances tests for TxtNode #141"

This update only reflect current status. It don't change anything.

- Rename TxtSyntax to `TxtAST` https://github.com/textlint/textlint/blob/master/docs/txtnode.d.ts
- Split `TxtNode`to `ParentNode`(block) and `TextNode`(inline)
- Define `value` property to TextNode
- Add Compliance tests tool https://github.com/textlint/textlint-ast-test
- Add note about https://github.com/wooorm/unist
- Add notice for Processor plugin creator


azu:
@homu r+

PR #181, opened 2016-03-28
fix(plugin): fix 3rd party plugin is not working

### LINKED ISSUE #180: Some plugins don't work
I'm using textlint v6.0.3. This version seems to not work plugins(More preceisely, non-embedded `Processor`).
I confirmed that it has occurred with `textlint-plugin-html` and my plugins in process of creation.

azu:
When load a plugin, then add available extensions like .html" to core.
But, `engine.availableExtensions` always is empty at  textlint v6.0.3.

fix #180

### DIFF
--- a/test/textlint-engine-test.js
+++ b/test/textlint-engine-test.js
@@ -122,13 +122,19 @@ describe("textlint-engine-test", function () {
                 assert(engine.ruleMap.getRule("example/example-rule") === ruleObject);
             });
         });
-        context("when load html plugin", function () {
+        context("when loading html plugin", function () {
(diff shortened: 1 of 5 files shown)


test/textlint-engine-test.js
```diff
@@ -122,6 +122,15 @@ describe("textlint-engine-test", function () {
                 assert(engine.ruleMap.getRule("example/example-rule") === ruleObject);
             });
         });
+        context("when load html plugin", function () {
```
azu: That is test for this bug.

azu:
@homu r+

PR #185, opened 2016-03-30
feat(core):  sort messages by line and column

### LINKED ISSUE #126: sort messages by line and column
We'll  sort message by line and column.

- https://github.com/textlint/textlint/blob/0ed7087810f6c266e60a73105ed752f8e4a3afd4/src/textlint-core.js#L99-L106

Add the sort like this:

```js
        // sort by line and column
        messages.sort(function(a, b) {
            var lineDiff = a.line - b.line;

            if (lineDiff === 0) {
                return a.column - b.column;
            } else {
                return lineDiff;
            }
        });
```

azu:
fix #126

### DIFF
--- a/src/fixer/fixer-processor.js
+++ b/src/fixer/fixer-processor.js
@@ -6,6 +6,7 @@ import FixerTask from "../task/fixer-task";
 import SourceCode from "../core/source-code";
 import SourceCodeFixer from "../fixer/source-code-fixer";
 import TaskRunner from "../task/task-runner";
+import sortMessages from "../util/sort-messages";
 export default class FixerProcessor {
     constructor(processor) {
         this.processor = processor;
@@ -82,7 +83,7 @@ export default class FixerProcessor {
             return {
                 filePath: resultFilePath,
                 output,
-                messages: originalMessages,
+                messages: sortMessages(originalMessages),
                 applyingMessages,
                 remainingMessages
             };
--- a/src/linter/linter-processor.js
+++ b/src/linter/linter-processor.js
@@ -3,6 +3,7 @@
 import assert from "assert";
 import LinterTask from "../task/linter-task";
 import TaskRunner from "../task/task-runner";
+import sortMessages from "../util/sort-messages";
 export default class LinterProcessor {
     /**
      * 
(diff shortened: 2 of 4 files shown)


azu:
FYI: https://github.com/1000ch/linter-textlint sometime display wrong order.
This pull request fix this issue.

azu:
@homu r+

io-monad:
Could this break the existing tests for rules? Requires major version up?

My rule plugin got failed after updated textlint to 5.6.0
https://github.com/io-monad/textlint-rule-general-novel-style-ja/pull/7

I'm using `textlint-tester` and the failed test case is:

```
        {
            text: "「こんにちは、世界。。。」",
            output: "「こんにちは、世界」",
            errors: [
                {
                    message: "句読点(。、)が閉じ括弧の直前に存在しています",
                    line: 1,
                    column: 12
                },
                {
                    message: "連続した句読点(。、)が使われています",
                    line: 1,
                    column: 10
                }
            ]
        },
```

azu:
@io-monad fmm.. I think that it is a implementation dependent(should be fixed), but the change break the test case.
I've reverted these commits and release 6.1.1.

https://github.com/textlint/textlint/commit/3651f9df61738d1ab49752885138cf436d8675e5
https://github.com/textlint/textlint/commit/106da2f7fa8ae927266df346cbbc89759417c6f7

io-monad:
@azu Yes, I agree that it should be fixed and definitely I will do. Or maybe textlint-tester can take care of the order of messages.

However, IMO changing the order of messages is not "in backwards-compatible manner" described in semver.

Anyway thanks for quick response!

azu:
It is better that One rule check One matter.
If you want to check multiple rule in a single module, I recommened that using preset system.

io-monad:
fmm, okay. I'll try splitting rules into local submodules like `textlint-rule-preset-JTF-style`. But I think plugin's name stays the same... (without officially recommended "textlint-rule-preset-" prefix). :sweat_smile:

I can say a rule preset may have tests for a conjunction of multiple sub-rules and may fail like this problem anyway, can't I? Or a rule preset should have no tests for conjunctions?

io-monad:
No, we don't need fix the test case for 6.1.0... :weary:

azu:
> :memo: It is better that One rule check One matter.

In other word, It is _Single Responsibility Principle_.

> I can say a rule preset may have tests for a conjunction of multiple sub-rules and may fail like this problem anyway, can't I? Or a rule preset should have no tests for conjunctions?

There are not official testing tool :bow: 
Can we need to add testing function like `runPreset` to [textlint-tester](https://github.com/textlint/textlint-tester "textlint-tester")?

azu:
We try again this  #218

Issue #47, opened 2015-10-23
Sharable configs


azu:
ESLint did http://eslint.org/docs/developer-guide/shareable-configs.html

1. First, `--config` support sharable config.
     - `--config @user/textlint-config`
2. <del>Second, extend sharable config by using `extend:"config"` in `.textlintrc` ?</del>

azu:
We don't like _extend config_. It it very difficult to understand.

azu:
I'm try to implement this :star:

azu:
Which is correct behavior?

1. First try to load Config module, Second try to load Config file.
1. First try to load Config filr, Second try to load Config module.

azu:
close by #186

azu:
https://github.com/textlint/textlint/releases/tag/6.2.0

PR #186, opened 2016-03-31
feat(config): support sharable config 

azu:
implement #47 
```
textlint --config <config-module>
textlint --config textlint-config-<name>
textlint --config @textlint/textlint-config
```


azu:
rc-loader related are not tested

azu:
@homu r+

PR #208, opened 2016-05-12
feat(ignore): add `{ ruleId }` option object to `shouldIgnore`

azu:
When the `ruleId` of option object is matched ruleId of error message, filter only the message,
Default is filter all message in the ignoring range.

## UseCase

https://github.com/textlint/textlint-rule-ignore-comment

```
<!-- textlint-disable ruleA -->
this is wrong
<!-- textlint-enable -->
```

Disable specific rule on the fly.


azu:
## Breaking Change(experimental)

`shouldIgnore(range)` only ignore messages of the `ruleId`.

It means that  `shouldIgnore(range)`  === `shouldIgnore(range, { ruleId : this.id })`.
It attempt to make safe by default.

"*" is danger because "*" affect to other rules by default.

Issue #211, opened 2016-05-22
`--fix` with ignoring rule


azu:
We have supported ignoring rule as experimental #193 

Currently, ignoring don't work with fixer.

This issue describe this bug. It is a bug of textlint-self.
https://github.com/textlint/textlint-rule-ignore-comments/issues/1

## Problem

Current, implementation of fixer.

> Parse text -> A fix rule -> re-parse text -> B fix rule ....

This design prefer much fixed than performance.

- https://github.com/textlint/textlint/blob/60633fe4530ee21e1da2d37386a568c5331f372c/src/fixer/fixer-processor.js#L36

**fixer and ignoring** 

Expected:

```
Parse text -> A fix rule -> re-parse text -> B fix rule ....
                  ^ + C ignoring rule                  ^ + C ignoring rule
```

Actual : 

> Parse text -> A fix rule -> re-parse text -> B fix rule -> re-parse text -> C ignoring rule ...

Not work ignoring.

It should be fixed

----

Edit(2016-05-26):

## Results after discussion

- Rename "ignoring rule" to "filter rule"
- Add "filters" filed support to textlint core
	- Same convention with "rules" field
- Rename "[textlint-rule-report-node-types](https://github.com/textlint/textlint-rule-report-node-types)" to "textlint-filter-rule-node-types". Is it better conventional name?
- Filter rule use `shouldIgnore(range, { ruleId })` API
  - https://github.com/textlint/textlint/blob/master/docs/rule.md
- textlint core still has actual filtering logic. - That is [message-filter.js](https://github.com/textlint/textlint/blob/master/src/shared/message-filter.js "message-filter.js")
  - [textlint-filter-plugin](https://github.com/textlint/textlint/issues/211#issuecomment-220991535) pattern is difficult.
  - We like simple.
- The order of filters doesn't affect result.
- textlint not allow (normal) rule to use `shouldIgnore`.

e.g.) The user want to use `textlint-filter-rule-node-types` and set following to  `.textlintrc` 

```json
{
   "filters": {
      "node-types": {
      	"nodeTypes": ["BlockQuote"]
      }
   }
}
```

See https://github.com/textlint/textlint/issues/211#issuecomment-221601511

azu:
How to know textlint that rule is ignoring rule?

Pattern A

ignoring rule should export `ignoring`

```js
module.exports = {
    ignoring: reporter
};
```

Pattern B

Add `ignore` to `textlintrc` 

```js
{
[...]

PR #213, opened 2016-05-28
refactor(messages): add MessageProcessManager

azu:
Added `MessageProcessManager` internal.
This manager process `TextLintMessage`s.
Main usecase filter messages.


src/messages/MessageProcessManager.js
```diff
@@ -0,0 +1,37 @@
+// LICENSE : MIT
+"use strict";
+export default class MessageProcessManager {
```
azu: 🆕

src/fixer/fixer-processor.js
```diff
@@ -51,7 +55,7 @@ export default class FixerProcessor {
 
                 return TaskRunner.process(task).then(messages => {
                     const result = postProcess(messages, sourceCode.filePath);
-                    result.messages = filterMessages(result.messages);
+                    result.messages = this.messageProcessManager.process(result.messages);
```
azu: fixer aplly filter, but it still has a issue #211

PR #217, opened 2016-05-31
feat(filter): add filter-duplicated-process

### LINKED ISSUE #209: Proposal: duplicated messages is ignored by default
## Summary

Make textlint preset more casual

## UseCase

The user use `preset-x` and `rule-a`.
`preset-x` contain `rule-a`, but also use `rule-a`.

```
{
    "rules": {
        "preset-x": {
        	// other many rules
            // ...
        	"rule-a": true
        },
        "rule-a": true
    }
}
```

In the case, `rule-a`'s error always is duplicated.

The user not want to see duplicated messages.

**Soluction A**

The user disable "rule-a" in the preset.

```
{
    "rules": {
        "preset-x": {
        	// other many rules
            // ...
        	"rule-a": false
        },
        "rule-a": true
    }
}
```

**Soluction B**

**textlint** ignore duplicated messages.

------

The above case is rare in current.
But incresing custom preset, this case will not be rare.

## Proposal

Duplicated messages is ignored by default.

**Duplicated messages** is that have same range and same message.
(Not depended on `ruleId`. `ruleId` is flexible property)

azu:
Duplicated messages is ignored by default.

**Duplicated messages** is that have same position information(`index`) and same `message`.

close #209


azu:
It contain a bit breanking change.
We will merge this as `--experimental`.

PR #218, opened 2016-06-01
feat(filter): re-enable sort messages by line and column

azu:
Add sort message feature as **experimental**

````
$ textlint --experimental
```

It contain a bit breaking change.
Previously we reject it on #185, but we try to implement it behind experimental flag.

refs #126


test/experimental/experimental-messages.js
```diff
@@ -0,0 +1,53 @@
+// LICENSE : MIT
+"use strict";
+/*
+    Main Task Run lint and work filter
+ */
+const assert = require("power-assert");
```
azu: This will be removed when textlint reach 7.0

Issue #219, opened 2016-06-07
Proposal: not load duplicated rule(ruleConfig)


azu:
We've alreadly ignore duplicated messages at #209 .

In advance, We want to ignore duplicated rule with same config.

```
{
    "rules": {
        "preset-x": {
            // other many rules
            // ...
            "rule-a": true
        },
        "rule-a": true
    }
}
```

If the rule `rule-a` and the config is `true`, load once.

Of cource, config is difference, should load `rule-a` twice.

```
{
    "rules": {
        "preset-x": {
            // other many rules
            // ...
            "rule-a": { "strict": true }
        },
        "rule-a": true
    }
}
```

PR #220, opened 2016-06-07
refactor(core): refactoring core-task/linter-task/fixer task

azu:
It aim to implement the proposal: not load duplicated rule(ruleConfig) #219

### DIFF
--- a/src/task/linter-task.js
+++ b/src/task/linter-task.js
@@ -40,8 +40,8 @@ export default class TextLintCoreTask extends CoreTask {
                 textLintConfig,
                 ruleConfig
             });
-            const ruleObject = this._getRuleObject(rule, ruleContext, ruleConfig);
-            this.addListenRule(ruleId, ruleObject);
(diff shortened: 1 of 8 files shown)


test/textlint/textlint-test.js
```diff
@@ -62,7 +62,7 @@ describe("textlint-test", function () {
                 };
                 textlint.setupRules({
                     "rule-name": rule
-                }, null);
```
azu: This is evil case.

src/task/linter-task.js
```diff
@@ -1,20 +1,61 @@
...
+        super();
+        this.config = config;
+        this.ruleCreatorSet = ruleCreatorSet;
+        this.filterRuleCreatorSet = filterRuleCreatorSet;
+        this.sourceCode = sourceCode;
+        this._setupRules();
+    }
+
+    start() {
+        this.startTraverser(this.sourceCode);
```
azu: delegate method.
We want to remove ti

PR #238, opened 2016-10-07
Fixed typos in README

rhysd:


azu:
@rhysd Thanks. Good Catch!

PR #249, opened 2016-11-21
docs(readme): Add VSCode extension information

taichi:


azu:
@taichi Thanks for Great Work!

PR #258, opened 2017-01-28
refactor(engine): use cosmiconfig instead of rc-loader

### LINKED ISSUE #39: rc-loader alternative?
We want to get an alternative of [MoOx/rc-loader](https://github.com/MoOx/rc-loader).
`rc` module is fragile, I think.
- for...in
  - https://github.com/dominictarr/rc/blob/d6ac3a3ac1555b25964e963fb949cd1d30b32c0f/lib/utils.js#L24
- `rc` use shabang
  - it break bundler like webpack.

RangeError: Maximum call stack size exceeded ?
## Alternative
- [spalger/rcloader](https://github.com/spalger/rcloader)

azu:
use [davidtheclark/cosmiconfig](https://github.com/davidtheclark/cosmiconfig "davidtheclark/cosmiconfig: Find and load configuration from a package.json property, rc file, or CommonJS module") insteadof rc-loader.
Maybe, this includes some *Breaking Change*.
Because, cosmiconfig have only async mode.

I have not motivation for using async config loading.
But, We want to drop to use rc-loader.
(rc-loader has some implicitly behavior)

We should weigh up using cosmicconfig and braking change.

fix #39


packages/textlint/src/engine/textlint-engine-core.js
```diff
@@ -122,7 +137,7 @@ new TextLintEngine({
      */
     loadPlugin(pluginName) {
         this.moduleLoader.loadPlugin(pluginName);
-        this._setupRules();
+        this._setupRules(this.config);
     }
```
azu: I think taht `load*` deprecated api work as well.
We should remove these api.

packages/textlint/test/mocha.opts
```diff
@@ -1,3 +1,4 @@
 --recursive
 --timeout 5000
 --compilers js:babel-register
+--require babel-polyfill
```
azu: It is needed for `async`/`await`

packages/textlint/src/engine/textlint-engine-core.js
```diff
@@ -84,23 +76,47 @@ export default class TextLintEngineCore {
...
+    load() {
+        return this.config.load().then((config) => {
+            this.config = config;
+            /**
+             * @type {TextLintCore}
+             */
+            this.textlint = new TextLintCore(config);
+            const cacheBaker = new CacheBaker(config);
+            if (config.cache) {
+                this.executeFileBackerManger.add(cacheBaker);
```
azu: Shold be independent.
it is worng that multiple `load()` then register twice.

azu:
This PR resolve following issue with webpack 2

```
ERROR in ./~/rc/index.js
Module parse failed: /Users/azu/.ghq/github.com/textlint/textlint.github.io/node_modules/rc/index.js Unexpected character '#' (1:0)
You may need an appropriate loader to handle this file type.
| #! /usr/bin/env node
| var cc   = require('./lib/utils')
| var join = require('path').join
 @ ./~/rc-loader/dist/index.js 10:10-23
 @ ./~/textlint/lib/config/config-loader.js
 @ ./~/textlint/lib/config/config.js
 @ ./~/textlint/lib/cli.js
 @ ./~/textlint/lib/index.js
 @ ./src/components/TextlintDemo/TextlintEditor.js
 @ ./src/components/TextlintDemo.js
 @ ./src/components/App.js
 @ ./src/index.js

```

azu:
This is closed by #262 
We use [azu/rc-config-loader](https://github.com/azu/rc-config-loader "azu/rc-config-loader: Load config from .{product}rc.{json,yml,js} file")

PR #278, opened 2017-05-05
feat(markdown-to-ast): update remark@7

azu:
BREAKING CHANGE: some parse result is changed.

It is correct, but we think that is a breacking change.
This change have possibilities that the user update patch-version textlint and fail test.

close https://github.com/textlint/markdown-to-ast/issues/14


azu:
I found a bit diffrence between remark 5 ... 7

```
- Paragraph is a `TxtParentNode` and type is Paragraph
    - have `children`, but not have `value`
*Emphasis* is a `TxtTextNode` and type is Emphasis
    - have `value`
"text" is a `TxtTextNode` and type is Str
    - have `value`
```

remark@7 http://bit.ly/2p47hnZ add `"\n"` text node:

```
    - have `children`, but not have `value`
*Emphasis* is a `TxtTextNode` and type is Emphasis
```

Added Node:

```
{
                  "type": "text",
                  "value": "\n",
                  "position": {
                    "start": {
                      "line": 4,
                      "column": 44,
                      "offset": 195
                    },
                    "end": {
                      "line": 5,
                      "column": 1,
                      "offset": 196
                    },
                    "indent": [
                      1
                    ]
                  }
                },
```

PR #283, opened 2017-05-07
Import textlint scripts

azu:
https://github.com/textlint/textlint-scripts


azu:
I think that textlint-scripts will not includes this monorepo.
because, it is very specilized application.

PR #292, opened 2017-05-14
Carve out textlint core logic into @textlint/kernel

### LINKED ISSUE #286: Carve out Core function to a core module
From #285 #270 

We already have `TextlintCore`, but it includes `fs` module.
We want to work textlint on browser.
To need carve out core core function that is not depended on `fs` to a module.

```js
import TextLintCoreCore from "@textlint/core"
const textlint = new TextLintCoreCore()
textlint.lint(text).then(() => {})
```

## Tasks

- [ ] Carve out Core into a module
- [ ] Need good name!

azu:
## Description

- Add 🆕 `@textlint/kernel` packages to `/pacakges`
- It has core logic of textlint
- It is not depended on Node.js feature. In other words, It work on browser.

## TODO

- [x] Update README
- [x] assert `options`
- [x] Add some tests for textlint-kernel
- [x] Remove unused dependencies from `textlint`
   - `yarn remove carrack traverse markdown-to-ast`

fix #286


azu:
We want to remove config in constructor.

- Plugin constructor
- Rule Context
- `--quite`

PR #315, opened 2017-10-06
chore(example): use --parallel --aggregate-output options for example test

azu:


azu:
close.

azu:
Can not pass tests...
Retry in the future.

Issue #304, opened 2017-09-23
textlint 9.0.0 Release Notes


azu:
# textlint 9.0.0 ChangeLog

###  textlint-kernel: Convert JavaScript to TypeScript #248 #301 

Convert [@textlint/kernel](https://github.com/textlint/textlint/tree/master/packages/textlint-kernel "@textlint/kernel") from JavaScript to Typescript.

If you instersting in TypeScript support, please see #248 .

Currently, Type definition is unstable. It means that break change in minor version.

###  Plugins can be initialized with plugin options #296 

Allow to specify following options.

```json
{
  "plugins": {
     "text": { "custom": "value" }
  }
}
```

For mode details, See [document of plugin](https://github.com/textlint/textlint/blob/master/docs/plugin.md#options).

###  textlint-tester: support "index" expected param #324 

Allow to write following `index` based testing.
```js
const TextLintTester = require("textlint-tester");
const tester = new TextLintTester();
tester.run("rule", rule, {
    invalid: [
        // index
        {
            text: "- [ ] string",
            errors: [
                {
                    message: "Found TODO: '- [ ] string'",
                    index: 2
                }
            ]
        }
    ]
});
```

### textlint: don't use error.code when catch error #323 

Fix to use correct exit code. Thanks to @buzztaiki

## 🔥 BREAKING CHANGE

These breaking change is for plugin/rule author.

###  Remove `context.config` #294 #295

You can use `context.getConfigBaseDir()` instead of `context.config`.

- `getConfigBaseDir()` return config base directory path that is the place of `.textlintrc`
    - e.g.) /path/to/dir/.textlintrc
    - `getConfigBaseDir()` return `"/path/to/dir/"`.

Related rules changes:

- [fix(rule): support textlint 8.2.0 by azu · Pull Request #15 · azu/textlint-rule-prh](https://github.com/azu/textlint-rule-prh/pull/15 "fix(rule): support textlint 8.2.0 by azu · Pull Request #15 · azu/textlint-rule-prh")
- [fix(textlint): support `config.getConfigBaseDir()` by azu · Pull Request #8 · azu/textlint-rule-eslint](https://github.com/azu/textlint-rule-eslint/pull/8 "fix(textlint): support `config.getConfigBaseDir()` by azu · Pull Request #8 · azu/textlint-rule-eslint")

###  Drop "rules" and "rulesConfig" from plugin #291 #313 

You can use `preset` feature instead of plugin.
The textlint plugin should focus on processor.

[...]

PR #323, opened 2017-10-27
fix(textlint): don't use error.code when catch error

buzztaiki:
error.code is a string, not a number.

https://nodejs.org/api/errors.html#errors_error_code


azu:
@buzztaiki Thanks for Pull Request!
This changes will be included in next (major) release. #304 

We will release next release at tomorrow.
You can try that beta release.

```
npm install textlint@9.0.0-beta.0
```

buzztaiki:
Thanks!

PR #382, opened 2017-12-14
fix(ast-node-types): Make TxtNode non-weak type

azu:
- `TxtNode` allow to set optional property
- Update txtnode.md
- Remove `TxtRootNode` from `@textlint/ast-node-types`

`@textlint/ast-node-types`'s scope is Abstract Syntax Tree.
It is inspiered by [unist](https://github.com/syntax-tree/unist "unist").
`@textlint/ast-node-types` does not include detailed Node interface like `StrNode`.

### DIFF
--- a/docs/txtnode.md
+++ b/docs/txtnode.md
@@ -17,25 +17,48 @@ Each node of the tree has same interface, is called `TxtNode`.
 `TxtNode` has these properties.
 
 ```typescript
+/**
+ * Basic TxtNode
+ * Probably, Real TxtNode implementation has more properties.
+ */
 interface TxtNode {
     type: string;
     raw: string;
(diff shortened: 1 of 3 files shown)


docs/txtnode.md
```diff
@@ -17,25 +17,48 @@ Each node of the tree has same interface, is called `TxtNode`.
...
-    range: [number, number]
-    loc: LineLocation;
+    range: TextNodeRange;
+    loc: TxtNodeLineLocation;
     // parent is runtime information
     // Not need in AST
+    // For example, top Root Node like `Document` has not parent.
     parent?: TxtNode;
+
+    [index: string]: any;
```
azu: OK: Optinal property

PR #394, opened 2017-12-18
feat(textlint-tester): support `inputPath` option

### LINKED ISSUE #390: textlint-tester: support `filePath`
Current, `textlint-tester` accept `text` property.
But, Some rule need to `filePath` property.
Because, `context.getFilePath()` return `undefined` if only pass `text` porperty.

UseCase:

- https://github.com/textlint-rule/textlint-rule-no-dead-link-fork/pull/3

azu:
#390 https://github.com/textlint-rule/textlint-rule-no-dead-link/issues/94

Add `inputPath` option for valid and invalid option.
We will need to add `outputPath` support.

fix #390

### DIFF
--- a/packages/textlint-tester/README.md
+++ b/packages/textlint-tester/README.md
@@ -15,13 +15,15 @@
 
 #### TextLintTester#run(ruleName, rule, {valid=[], invalid=[]})
 
+##### valid object
+
 - `{string} ruleName` ruleName is a name of the rule.
 - `{Function} rule` rule is the exported function of the rule.
 - `{string[]|object[]} valid` valid is an array of text which should be passed.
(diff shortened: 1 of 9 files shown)


azu:
I thinking about this, `inputPath` to be `filePath`
Current `filePath` is a real path. It exists actually.
(It lint by `lintFile`)

But, we could treat `text` and pseudo-`filePath` in the future.
The `filePath` is pseudo file path.
(It will be linted by kernel.lintText with option)

azu:
I decide to use `inputPath`.

azu:
Ah, should expand following 

```
===Text===:
${text}
```

text is [object Object] is bad.

PR #397, opened 2017-12-18
Validate config utf8

### LINKED ISSUE #298: .textlintrc should be encoded as UTF-8
Currently, textlint throw error when `.textlintrc` file is encoded as Shift-JIS.
We should check that the `.textlintrc` is UTF-8.

## Proposal

If `.textlintrc` is not encoded by UTF-8, show error mesage in console.

## Modules

- [utf-8-validate](https://www.npmjs.com/package/utf-8-validate)
  - It is native module
  - Previoulsly, it is introduced in #397, but it is reverted in #509 

Edit: Re-open by #509

Leko:
This PR closes #298

- Throw an error when config file is not encoded in UTF8
- Add tests


packages/textlint/src/config/config-loader.ts
```diff
@@ -2,6 +2,8 @@ import { TextLintModuleResolver } from "../engine/textlint-module-resolver";
 
 // LICENSE : MIT
 "use strict";
+const fs = require("fs");
```
azu: It is better in TypeScript.
`@types/node` support Node.js Core module typing.
```ts
import * as fs from "fs"
```
It make it typing.

PR #403, opened 2017-12-19
fix(textlint-formatter): run all tests

### LINKED ISSUE #402: No formatter test will be executed
Hi, I found issues while implementing possible fix for #400 locally.

v10.0.0

N/A

Testing `textlint-formatter` by `yarn run test`

Formatters tests, including tests under [`test/formatters/`](https://github.com/textlint/textlint/tree/c0693b79a48ba84620dc57235667f3d6219e174f/packages/textlint-formatter/test/formatters) directory, will be executed.

Only [`textlint-formatter-test.js`](https://github.com/textlint/textlint/blob/c0693b79a48ba84620dc57235667f3d6219e174f/packages/textlint-formatter/test/textlint-formatter-test.js) was executed.

This seems to be due to glob mismatch in the [`package.json`](https://github.com/textlint/textlint/blob/c0693b79a48ba84620dc57235667f3d6219e174f/packages/textlint-formatter/package.json#L19). Solution could be as follows. Could you give me your thoughts?

```diff
diff --git a/packages/textlint-formatter/package.json b/packages/textlint-formatter/package.json
index 4c1114b..ac9de7d 100644
--- a/packages/textlint-formatter/package.json
+++ b/packages/textlint-formatter/package.json
@@ -16,7 +16,7 @@
     "bin/"
   ],
   "scripts": {
-    "test": "mocha test/*.js"
+    "test": "mocha test/**/*.js"
   },
   "directories": {
     "test": "test/"
```

...and you will get 14 failing tests for `stylish` and `table` formatters 😣

0x6b:
Correct mocha test glob pattern to run all tests. Closes #402.

### DIFF
--- a/packages/textlint-formatter/package.json
+++ b/packages/textlint-formatter/package.json
@@ -16,7 +16,7 @@
     "bin/"
   ],
   "scripts": {
-    "test": "mocha test/*.js"
+    "test": "mocha test/**/*.js"
   },
   "directories": {
     "test": "test/"
--- a/packages/textlint-formatter/package.json
+++ b/packages/textlint-formatter/package.json
@@ -16,7 +16,7 @@
     "bin/"
   ],
   "scripts": {
-    "test": "mocha test/**/*.js"
+    "test": "mocha --no-colors test/**/*.js"
   },
   "directories": {
     "test": "test/"


packages/textlint-formatter/package.json
```diff
@@ -16,7 +16,7 @@
     "bin/"
   ],
   "scripts": {
-    "test": "mocha test/*.js"
+    "test": "mocha test/**/*.js"
```
azu: Can you try --no-color option?

https://github.com/textlint/textlint/issues/402#issuecomment-352808734

0x6b: Run test from top directory without `--no-color` have passed successfully.

0x6b: Should I add `--no-color` option to the `package.json`, or leave it as it is (at this moment, and fix test cases later)?

azu: Please add `--no-color` option.
master branch should be always passed tests in any environment.

0x6b:
Thank you!

PR #407, opened 2017-12-21
Upgrade mocha 4

Leko:
This PR closes #405 

- Update mocha `^4.0.1`
- Use `--require` instead of `--compilers`
- Use glob pattern instead of `--recursive`

### Other changes
- Use `@textlint/ast-node-types` instead of relative path

PR #410, opened 2017-12-23
refactor(textlint-formatter): remove reference to always empty string

### LINKED ISSUE #400: source property in TextLintMessage class

v10.0.0

N/A

I would like to fix issue #399. Before adding new function(`getFormatterList`), I tried to convert `textlint-formatter` to TypeScript first as same as `textlint-fixer-formatter`.

I found no `property` in `TextLintMessage`. Which is expected?

- [`jslint-xml`](https://github.com/textlint/textlint/blob/5d72ce3270727e729416bfda0c4b6c72e611b18a/packages/textlint-formatter/test/formatters/jslint-xml.js) formatter [uses](https://github.com/textlint/textlint/blob/44ae6153e4035fd765656d0c3d4154fc72fc2429/packages/textlint-formatter/lib/formatters/jslint-xml.js#L28) `source` property ([test case](https://github.com/textlint/textlint/blob/41cae972db08731b797592b24dcd845c0138bec4/packages/textlint-formatter/test/formatters/jslint-xml.js#L31)), while
- [`TextLintMessage`](https://github.com/textlint/textlint/blob/44ae6153e4035fd765656d0c3d4154fc72fc2429/packages/%40textlint/kernel/src/textlint-kernel-interface.ts#L171) class does not have such property.

I think I'm missing something. Appreciate your comments. Thanks!

0x6b:
Modify test case as well. Closes #400.

Please note that commit 72136e3 (#407) reverts commit 1548585 (#403) and causes failed test in local.

```sh
cd packages/textlint-formatter
yarn run test # fail
yarn run test --no-color # success
```

### DIFF
--- a/packages/textlint-formatter/lib/formatters/jslint-xml.js
+++ b/packages/textlint-formatter/lib/formatters/jslint-xml.js
@@ -25,7 +25,9 @@ module.exports = function(results) {
         messages.forEach(function(message) {
             output += "<issue line=\"" + message.line + "\" " +
                 "char=\"" + message.column + "\" " +
-                "evidence=\"" + lodash.escape(message.source || "") + "\" " +
+                // TODO: evidence is always empty string
+                // See: https://github.com/textlint/textlint/issues/400
+                "evidence=\"\" " +
                 "reason=\"" + lodash.escape(message.message || "") +
                 (message.ruleId ? " (" + message.ruleId + ")" : "") + "\" />";
         });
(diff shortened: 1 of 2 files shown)


azu:
@0x6b Thanks! 

> Please note that commit 72136e3 (#407) reverts commit 1548585 (#403) and causes failed test in local.

Oh, I've missed it. /cc @Leko

Issue #399, opened 2017-12-18
formater: Add `getFormatterList` function


azu:
[textlint-formatter](https://github.com/textlint/textlint/tree/master/packages/textlint-formatter "textlint-formatter") and [textlint-fixer-formatter](https://github.com/textlint/textlint/tree/master/packages/textlint-fixer-formatter "textlint-fixer-formatter") does not export own formatter list like `"stylish"`, `"compats"` etc...

It need to resolve #85

- [show available formatter in help · Issue #85 · textlint/textlint](https://github.com/textlint/textlint/issues/85 "show available formatter in help · Issue #85 · textlint/textlint")

## Tasks

- [x] Add `getFormatterList` function to [textlint-formatter](https://github.com/textlint/textlint/tree/master/packages/textlint-formatter "textlint-formatter")  #415 
- [x] Add `getFormatterList` function to [textlint-fixer-formatter](https://github.com/textlint/textlint/tree/master/packages/textlint-fixer-formatter "textlint-fixer-formatter") #412 

`getFormatterList()` returns an array of `{ name: "formatter name"}` object.

For example, [textlint-formatter](https://github.com/textlint/textlint/tree/master/packages/textlint-formatter "textlint-formatter")'s `getFormatterList()`  returns following array of object:

```js
[{
  name: "stylish"
}, {
  name: "compact"
}]
```

- ... [textlint-formatter/lib/formatters/](https://github.com/textlint/textlint/tree/master/packages/textlint-formatter/lib/formatters "textlint-formatter/lib/formatters/") file names ...

Probably, [fs.readdirSync](https://nodejs.org/api/fs.html#fs_fs_readdirsync_path_options "fs.readdirSync") help this implementation.

azu:
I've edit `getFormatterList()` result value.

```diff
- return an array of formatter names
+ return an array of `{ name: "formatter name"}` object
```

It aim to get extensibility.

azu:
👏 Closed by @0x6b PR #415 #412

PR #412, opened 2017-12-23
feat(textlint-fixer-formatter): add getFormatterList function

0x6b:
Hi, could you review my preliminary `getFormatterList` implementation for `textlint-fixer-formatter`(#399)?

### DIFF
--- a/packages/textlint-fixer-formatter/test/textlint-fixer-formatter-test.js
+++ b/packages/textlint-fixer-formatter/test/textlint-fixer-formatter-test.js
@@ -1,0 +1,19 @@
+// LICENSE : MIT
+'use strict'
+var path   = require('path')
+var assert = require('power-assert')
+import { getFormatterList } from 'textlint-fixer-formatter'
+
+describe('textlint-fixer-formatter-test', function () {
+  describe('getFormatterList', function () {
+    it('should return list of formatter(s)', function () {
+      assert(getFormatterList(), [
+          {name: 'compact'},
+          {name: 'diff'},
+          {name: 'json'},
+          {name: 'stylish'}
+        ]
+      )
+    })
+  })
+})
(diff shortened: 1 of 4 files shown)


packages/textlint-fixer-formatter/test/textlint-fixer-formatter-test.js
```diff
@@ -0,0 +1,19 @@
+// LICENSE : MIT
```
azu: Can you convert it to TypeScript?

It is good that adding new test file as TypeScript.

0x6b: It makes sense. Thanks. I also found that only `assert` is not meaningful. Trying to fix.

azu:
@0x6b Can you forget to apply `prettier` formatter?

    # project root
    yarn run prettier

https://github.com/textlint/textlint/blob/daeafe18bf19b7b206d1d963b9195aeb4c8a8ea8/package.json#L30
textlint install `precommit` hook when `yarn install` at first time.
 So, we can apply prettier automatically, but it was not applied... mystery.

0x6b:
Mysterious. Maybe husky :dog: haven't work as expected recently. I will check my setup but, in the meantime, apply prettier manually

azu:
Thanks for fixing!

0x6b:
Really appreciate your multiple reviews 😄 Please mark one of the tasks at #399 as done. Thank you!

0x6b:
> Mysterious. Maybe husky 🐶 haven't work as expected recently. I will check my setup but, in the meantime, apply prettier manually

I set up my own hooks at `~/.gitconfig#core.hookspath` and it prevents Husky from running (see https://github.com/typicode/husky/issues/81#issuecomment-282540588). Will open an issue for updating [CONTRIBUTING.md](https://github.com/textlint/textlint/blob/master/.github/CONTRIBUTING.md) about this.

PR #413, opened 2017-12-23
Convert textlint-formatter to TypeScript

0x6b:
As I said at https://github.com/textlint/textlint/issues/400#issue-282950494, I would like to fix issue #399. Before adding `getFormatterList` to `textlint-formatter`, I tried to convert the module to TypeScript. Please review my pull request. Since I'm relatively new to TypeScript, your comments would be appreciated.

Closes #413

PR #414, opened 2017-12-23
fix(monorepo): fix TypeScript module resolution in monorepo

azu:
- Use `index.ts` **Important**
- Extends tsconfig.base.json

## Problem

#413 Cause following build-time error:

```
 ../@textlint/kernel/src/core/BaseRuleContext.ts(1,39): error TS2307: Cannot find module '@textlint/ast-node-types'.
lerna ERR! ../@textlint/kernel/src/core/filter-rule-context.ts(4,25): error TS2307: Cannot find module '@textlint/ast-node-types'.
```

This error is caused by build dependency.
`textlint-formatter` -> depend on -> `@textlint/kernel` -> depend on -> `@textlint/ast-node-types`
When building `textlint-formatter`, `@textlint/kernel`'s `lib/` is not built yet.
It cause reference error like "TS2307: Cannot find module"

But, we already have `path` mapping for avoiding this error.
It make `textlint-formatter` refer to `src/`  instead of `lib/`.
https://github.com/textlint/textlint/blob/37bccdb9a91382c00deb914e4e7195eba9700a94/tsconfig.base.json#L4-L11

- [Module Resolution · TypeScript](https://www.typescriptlang.org/docs/handbook/module-resolution.html "Module Resolution · TypeScript")
- [Quramy/lerna-yarn-workspaces-example: How to build TypeScript mono-repo project with yarn and lerna](https://github.com/Quramy/lerna-yarn-workspaces-example "Quramy/lerna-yarn-workspaces-example: How to build TypeScript mono-repo project with yarn and lerna")

The `paths` mapping is  incompleteness.

```ts
    "baseUrl": ".",
    "paths": {
      "@textlint/*": [
        "./packages/@textlint/*/src"
      ],
      "*": [
        "./packages/*/src"
      ]
    },
```

This `paths` mapping means that `@textlint/ast-node-types` to `packages/@textlint/ast-node-types/src`

```js
require("@textlint/ast-node-types")
// ===
require("packages/@textlint/ast-node-types/src")
```

This is **incompleteness**.

Because, `require("packages/@textlint/ast-node-types/src")` refer to `require("packages/@textlint/ast-node-types/src/index.ts")`.
But,  `packages/@textlint/ast-node-types/src/index.ts` is not found.
As as result, `require("@textlint/ast-node-types")` can not resoluve `paths`.

## Resolution

Put `index.ts` to `packages/@textlint/ast-node-types/src/index.ts`.
As a result,  `require("@textlint/ast-node-types")` can be resolved as  `packages/@textlint/ast-node-types/src/index.ts`.

It looks like workaround.

We want to get more explicit solution.


packages/@textlint/ast-node-types/package.json
```diff
@@ -2,8 +2,8 @@
   "name": "@textlint/ast-node-types",
   "version": "3.0.0",
   "description": "textlint AST node type definition.",
-  "main": "./lib/TextLintASTNodeTypes.js",
-  "types": "./lib/TextLintASTNodeTypes.d.ts",
+  "main": "./lib/index.js",
```
azu: implicit `index` is very important meaning.

azu:
A fundamental point is that TypeScript's `paths`  does resolve path **without** `main` field.

Issue #369, opened 2017-12-12
Convert tests to TypeScript


azu:
We have conveted source codes of [@textlint/kernel](https://github.com/textlint/textlint/tree/master/packages/%40textlint/kernel "@textlint/kernel") and [textlint](https://github.com/textlint/textlint/tree/master/packages/textlint "textlint") to TypeScript.
But, test codes is still JavaScript.

We want to convert to(or add) TypeScript test for it.

## Main purposes

- Add new TypeScript test code
  - It make the app stronge 
- Convert JavaScript test code to TypeScript code if it is possible
  - Some test is very hacky
  - It is not must

TypeScript test code help to find wrong type interface.

For example, current `textlint-engine` and `textfix-engine` is very strange implementation.
In other words, this implementaion is very hackly way.
Probably, It will be needed to refactor for passing typescript test code.

- https://github.com/textlint/textlint/blob/da23f7171912ccad3174f3028b2790e8f099e26f/packages/textlint/src/textlint-engine.ts#L61
- https://github.com/textlint/textlint/blob/da23f7171912ccad3174f3028b2790e8f099e26f/packages/textlint/src/textfix-engine.ts#L60

Adding or Converting test is that blame it 🔥 

## Priority

We want to get started with converting Public API testing.
Public API means that textlint and @textlint/kernel export modules.

- [textlint/src/index.ts](https://github.com/textlint/textlint/blob/master/packages/textlint/src/index.ts "textlint/src/index.ts")
- [@textlint/kernel/src/index.ts](https://github.com/textlint/textlint/blob/master/packages/%40textlint/kernel/src/index.ts "@textlint/kernel/src/index.ts")

Convert Public API testing:

**[textlint](https://github.com/textlint/textlint/tree/master/packages/textlint "textlint")**:

- [x] TextLintCore
- [x] TextFixEngine
- [x] TextLintEngine
- [x] textlint
- [x] cli

**[@textlint/kernel](https://github.com/textlint/textlint/tree/master/packages/%40textlint/kernel "@textlint/kernel")**:

- [x] [textlint-kernel-test.ts](https://github.com/textlint/textlint/blob/master/packages/%40textlint/kernel/test/textlint-kernel-test.ts "textlint-kernel-test.ts")
  - But more tests for various options

[...]

PR #420, opened 2017-12-26
refactor(ast-node-types): convert ASTNodeTypes test to TypeScript

0x6b:
- Convert test case to TypeScript as part of #369
- Add type definition to `ASTNodeTypes`
- Correct `main` and `types` properties in `package.json` to prevent
  module resolving issue at `@textlint/kernel`
- Add `clean` package script and dev dependency for build stability


packages/@textlint/ast-node-types/src/index.ts
```diff
@@ -1,11 +1,12 @@
 // MIT © 2017 azu
 "use strict";
+
 /**
  * AST Node types list on TxtNode.
  * Constant value of types
  * @see https://github.com/textlint/textlint/blob/master/docs/txtnode.md
  */
-export const ASTNodeTypes = {
+export const ASTNodeTypes: { [key: string]: string } = {
```
azu: 📝 We can use String enum that is introduced in TypeScript 2.4.
It will be other PR. (I need more research)

```ts
export enum ASTNodeTypes {
    Document = "Document",
    Paragraph = "Paragraph",
    BlockQuote = "BlockQuote",
    ListItem = "ListItem",
    List = "List",
    Header = "Header",
    CodeBlock = "CodeBlock",
    HtmlBlock = "HtmlBlock",
    ReferenceDef = "ReferenceDef",
    HorizontalRule = "HorizontalRule",
    Comment = "Comment",
    // inline
    Str = "Str",
    Break = "Break", // well-known Hard Break
    Emphasis = "Emphasis",
    Strong = "Strong",
    Html = "Html",
    Link = "Link",
    Image = "Image",
    Code = "Code",
    Delete = "Delete"
};
```

packages/@textlint/ast-node-types/package.json
```diff
@@ -2,8 +2,8 @@
   "name": "@textlint/ast-node-types",
   "version": "3.0.1",
   "description": "textlint AST node type definition.",
-  "main": "./lib/index.js",
-  "types": "./lib/index.d.ts",
+  "main": "./lib/src/index.js",
+  "types": "./lib/src/index.d.ts",
```
azu: > Correct main and types properties in package.json to prevent
> module resolving issue at @textlint/kernel

Probably, This changes is caused by adding `test/TextLintASTNodeTypes-test.ts`
I've noticed that [@textlint/ast-node-types/tsconfig.json](https://github.com/textlint/textlint/blob/master/packages/%40textlint/ast-node-types/tsconfig.json "@textlint/ast-node-types/tsconfig.json") does not specified `includes` target.

When no have `includes` setting in tsconfig.json, `tsc` convert all `.ts` files.
In other word, current config is `includes: ["*"]`.

> If the "files" and "include" are both left unspecified, the compiler defaults to including all TypeScript (.ts, .d.ts and .tsx)
> https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

Alternative solution:

1. Revert this lines
2. Add `includes` to [@textlint/ast-node-types/tsconfig.json](https://github.com/textlint/textlint/blob/master/packages/%40textlint/ast-node-types/tsconfig.json "@textlint/ast-node-types/tsconfig.json")

```js
{
  "extends": "../../tsconfig.base.json",
  "compilerOptions": {
    "outDir": "./lib/"
  },
  "include": [
    "./src/**/*"
  ],
  "exclude": [
    ".git",
    "node_modules"
  ]
}
```

I prefer latter solution.

@0x6b How do you think?

packages/@textlint/ast-node-types/package.json
```diff
@@ -13,6 +13,7 @@
     "url": "https://github.com/textlint/textlint.git"
   },
   "scripts": {
+    "clean": "rimraf lib/",
```
azu: I've created a issue about `clean`

https://github.com/textlint/textlint/issues/421

packages/@textlint/ast-node-types/src/index.ts
azu: I've filed it as issue.
https://github.com/textlint/textlint/issues/422

packages/@textlint/ast-node-types/package.json
0x6b: @azu, thank you. I have a lot to learn! Since I left the `tsconfig.json` as is, I had to modify `main` and `types` properties, but felt uncomfortable a bit 😒 for the directory structure (hey, `src` just under `lib`...?) I will push another commits which implement your solution.

0x6b:
My first merge to `textlint:master` 👏 . @azu, thank you for your review.

PR #429, opened 2017-12-28
refactor(textlint-fixer-formatter): make `getFormatterList()` return type explicit 

azu:
## Breaking Change

- Renamed `getFormatterList` to `getFixerFormatterList`

## Refactor

- make `getFormatterList()` function return type explicit
  - Define `FixerFormatterDetail[]`
  - `textlint-formatter` already support `FormatterDetail`
- textlint use the `FormatterDetail` and `FixerFormatterDetail`

### DIFF
--- a/packages/textlint-fixer-formatter/src/index.ts
+++ b/packages/textlint-fixer-formatter/src/index.ts
@@ -46,7 +46,11 @@ ${ex}`);
     };
 }
 
-export function getFormatterList() {
+export interface FormatterDetail {
+    name: string;
+}
+
+export function getFormatterList(): FormatterDetail[] {
     return fs
         .readdirSync(path.join(__dirname, "formatters"))
         .filter((file: string) => {
--- a/packages/textlint-fixer-formatter/src/index.ts
+++ b/packages/textlint-fixer-formatter/src/index.ts
@@ -49,7 +49,7 @@ export interface FixerFormatterDetail {
     name: string;
 }
 
-export function getFormatterList(): FixerFormatterDetail[] {
+export function getFixerFormatterList(): FixerFormatterDetail[] {
     return fs
         .readdirSync(path.join(__dirname, "formatters"))
         .filter((file: string) => {
--- a/packages/textlint-fixer-formatter/src/index.ts
+++ b/packages/textlint-fixer-formatter/src/index.ts
@@ -21,7 +21,6 @@ export function createFormatter(formatterConfig: FormatterConfig) {
     } else if (fs.existsSync(path.resolve(process.cwd(), formatterName))) {
         formatterPath = path.resolve(process.cwd(), formatterName);
     } else {
-        // FIXME: Move textfix-formatter to pacakges/
         if (isFile(`${path.join(__dirname, "formatters/", formatterName)}.js`)) {
             formatterPath = `${path.join(__dirname, "formatters/", formatterName)}.js`;
         } else if (isFile(`${path.join(__dirname, "formatters/", formatterName)}.ts`)) {
@@ -46,11 +45,11 @@ ${ex}`);
     };
 }
 
-export interface FormatterDetail {
+export interface FixerFormatterDetail {
     name: string;
 }
 
-export function getFormatterList(): FormatterDetail[] {
+export function getFormatterList(): FixerFormatterDetail[] {
     return fs
         .readdirSync(path.join(__dirname, "formatters"))
         .filter((file: string) => {
--- a/packages/textlint/src/options.ts
+++ b/packages/textlint/src/options.ts
@@ -1,11 +1,11 @@
 // LICENSE : MIT
 "use strict";
-import { getFormatterList } from "textlint-formatter";
+import { getFormatterList, FormatterDetail } from "textlint-formatter";
 import { getFormatterList as getFixerFormatterList } from "textlint-fixer-formatter";
 
 const optionator = require("optionator");
 
-const concatFormatterList = (formatterList: { name: string }[]) => {
+const concatFormatterList = (formatterList: FormatterDetail[]) => {
     return formatterList
         .map(formatter => {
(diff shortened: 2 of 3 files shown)


packages/textlint-fixer-formatter/src/index.ts
```diff
@@ -46,7 +46,11 @@ ${ex}`);
     };
 }
 
-export function getFormatterList() {
+export interface FormatterDetail {
+    name: string;
+}
+
+export function getFormatterList(): FormatterDetail[] {
```
azu: Public API should have explicit interface.

0x6b: @azu, it looks ok for me. But `textlint-formatter` already has `FormatterDetail` interface. To avoid duplication, how about following changes? (the interface is small, though)

```diff
diff --git a/packages/textlint-fixer-formatter/src/index.ts b/packages/textlint-fixer-formatter/src/index.ts
index d8cda30..f0db498 100644
--- a/packages/textlint-fixer-formatter/src/index.ts
+++ b/packages/textlint-fixer-formatter/src/index.ts
@@ -1,6 +1,7 @@
 // LICENSE : MIT
 "use strict";
 import { TextlintFixResult } from "@textlint/kernel";
+import { FormatterDetail } from "textlint-formatter";

 const fs = require("fs");
 const path = require("path");
@@ -46,7 +47,7 @@ ${ex}`);
     };
 }

-export function getFormatterList() {
+export function getFormatterList(): FormatterDetail[] {
     return fs
         .readdirSync(path.join(__dirname, "formatters"))
         .filter((file: string) => {
```

 This will introduce additional dependency to `textlint-formatter`, but since `textlint-fixer-formatter`

azu: I think that the duplication is OK.
Because, `textlint-fixer-formatter`'s `FormatterDetail` can be difference with `textlint-formatter`s `FormatterDetail`.
(Also,  `textlint-fixer-formatter` does not depnended on `textlint-formatter` in semantically)

azu: I think that rename `FormatterDetail` to `FixerFormatterDetail` is reasone.

@0x6b Do you think about it ?

0x6b: @azu, thank you.

> Because, `textlint-fixer-formatter`'s `FormatterDetail` can be difference with `textlint-formatter`s `FormatterDetail`.

This makes sense for me, and renaming new interface to`FixerFormatterDetail` LGTM.

0x6b: But `packages/textlint/src/options.ts` implementation might be a bit awful...?

azu:
fixed

packages/textlint-fixer-formatter/src/index.ts
```diff
@@ -46,7 +45,11 @@ ${ex}`);
     };
 }
 
-export function getFormatterList() {
+export interface FixerFormatterDetail {
+    name: string;
+}
+
+export function getFormatterList(): FixerFormatterDetail[] {
```
0x6b: For consistency, `getFormatterList()` could be renamed to `getFixerFormatterDetail()`. Could you let me know your thoughts.

package | API | returns
-|-|-
`textlint-formatter` | `getFormatterList()` | `FormatterDetail[]`
`textlint-fixer-formatter` | `getFormatterList()` or `getFixerFormatterList()`? | `FixerFormatterDetail[]`

packages/textlint/src/options.ts
```diff
@@ -1,11 +1,11 @@
 // LICENSE : MIT
 "use strict";
-import { getFormatterList } from "textlint-formatter";
+import { getFormatterList, FormatterDetail } from "textlint-formatter";
 import { getFormatterList as getFixerFormatterList } from "textlint-fixer-formatter";
 
 const optionator = require("optionator");
 
-const concatFormatterList = (formatterList: { name: string }[]) => {
+const concatFormatterList = (formatterList: FormatterDetail[]) => {
```
0x6b: `concatFormatterList()` should accept both `FormatterDetail[]` and `FixerFormatterDetail[]`? e.g.

```diff
diff --git a/packages/textlint/src/options.ts b/packages/textlint/src/options.ts
index 33a1017..62a0e0d 100644
--- a/packages/textlint/src/options.ts
+++ b/packages/textlint/src/options.ts
@@ -1,11 +1,11 @@
 // LICENSE : MIT
 "use strict";
-import { getFormatterList } from "textlint-formatter";
-import { getFormatterList as getFixerFormatterList } from "textlint-fixer-formatter";
+import { getFormatterList, FormatterDetail } from "textlint-formatter";
+import { getFormatterList as getFixerFormatterList, FixerFormatterDetail } from "textlint-fixer-formatter";

 const optionator = require("optionator");

-const concatFormatterList = (formatterList: { name: string }[]) => {
+const concatFormatterList = (formatterList: FormatterDetail[] | FixerFormatterDetail[]) => {
     return formatterList
         .map(formatter => {
             return formatter.name;
```

packages/textlint-fixer-formatter/src/index.ts
azu: > getFixerFormatterList

Reasonable.
I've fixed it.

packages/textlint/src/options.ts
azu: Nice catch!

Fixed in https://github.com/textlint/textlint/pull/429/commits/c27bacc98420ca48d83dd44038749b3a14c356e5

0x6b:
LGTM

packages/textlint-fixer-formatter/src/index.ts
0x6b: Not awful as I thought first. Union type is cool.

PR #436, opened 2017-12-31
fix(feature-flag): Update MapLike@2

azu:
Use map-like@2


yarn.lock
```diff
@@ -4353,10 +4353,6 @@ map-cache@^0.2.0:
   version "0.2.2"
   resolved "https://registry.yarnpkg.com/map-cache/-/map-cache-0.2.2.tgz#c32abd0bd6525d9b051645bb4f26ac5dc98a0dbf"
 
-map-like@^1.1.2:
```
azu: It cause conflict Map Like 1.x and 2.x.

This PR resolve this issue.

Issue #386, opened 2017-12-15
Add `yarn check --integrity` in `yarn test`


azu:
Currently, The difference between `yarn.lock` and pacakge.json is ignored.
(CI doesn't fail)

We can check the difference by [yarn check --integrity](https://yarnpkg.com/lang/en/docs/cli/check/ "yarn check") command.

## Workflow

Does check the difference between `yarn.lock` and `package.json` dependencies.

```
yarn check --integrity
```

If the command is failed, Run `yarn run bootstrap` that update `yarn.lock`.

## Proposal

Add `yarn-check` command to `pacakge.json`

```
"test": "npm-run-all --parallel yarn-check lint test:packages",
// ...   
"yarn-check": "yarn check --integrity || echo 'Please run yarn bootstrap'"
```

**Problem**

This proposal has a problem.
It is always to be exit code `0` and, maybe, it is not compatible with cross platform.
We want to improve this.

```
✈ yarn check --integrity || echo 'Please run yarn bootstrap'
yarn check v1.3.2
warning Integrity check: Lock files don't match
error Integrity check failed
error Found 1 errors.
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.
Please run yarn bootstrap

~/.ghq/github.com/textlint/textlint master*
✈ echo $?
0
```

Leko:
I think `()` would be better to fallback multiple commands.

```
"yarn-check": "yarn check --integrity || (echo 'Please run yarn bootstrap' && exit 1)"
```

`yarn-check` will output the error message and keep exit code.

azu:
Reopend by #438 

`yarn-check` in `precommit` or `prepush` is conflict with `lerna publish` workflow

- https://github.com/lerna/lerna#publish

lerna modified `package.json` and commit, and `yarn-check` is failed.
Additionaly, we should `yarn bootstrap` after `lerna publish`.
Because, `lerna publish` only modified `package.json`.
It does not `yarn.lock` file.

azu:
> yarn-check in precommit or prepush is conflict with lerna publish workflow

This was lerna bug.
It is fixed in 2.8.0

- [Release v2.8.0 · lerna/lerna](https://github.com/lerna/lerna/releases/tag/v2.8.0 "Release v2.8.0 · lerna/lerna")
[...]

PR #438, opened 2017-12-31
[Test] New release

azu:
- It is test.
- #386 is conflict with `lerna publish` workflow 😢 
  - https://github.com/lerna/lerna#publish
  - I've removed `yarn-check` in precommit hook


azu:
We will maintenance release/publish flow.

PR #444, opened 2018-01-03
fix(kernel): fix return type of `SourceCode#getSource`

azu:
`SourceCode#getSource` never return `null`.
Because, `this.text` always has value.

### DIFF
--- a/packages/@textlint/kernel/src/core/source-code.ts
+++ b/packages/@textlint/kernel/src/core/source-code.ts
@@ -36,6 +36,7 @@ export interface SourceCodeLocation {
 }
 
 export type SourceCodeRange = [number, number];
+export type SourceCodeArgs = { text: string; ast: TxtNode; ext: string; filePath?: string };
 /**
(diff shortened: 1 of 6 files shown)


packages/@textlint/kernel/src/core/source-code.ts
```diff
@@ -92,13 +93,10 @@ export default class SourceCode {
      * @param {TxtNode=} node The AST node to get the text for.
      * @param {int=} beforeCount The number of characters before the node to retrieve.
      * @param {int=} afterCount The number of characters after the node to retrieve.
-     * @returns {string|null} The text representing the AST node.
+     * @returns {string} The text representing the AST node.
      */
-    getSource(node: TxtNode, beforeCount?: number, afterCount?: number) {
+    getSource(node?: TxtNode, beforeCount?: number, afterCount?: number): string {
         const currentText = this.text;
-        if (currentText == null) {
```
azu: `this.text`  is not null.

Issue #455, opened 2018-01-11
Move core modules to @textlint namespace


azu:
npm recommened to use scoped.

> If you are prevented from publishing a package with a name too similar to an existing package, the easiest way to find a unique name is to use your scope. You can scope packages to your npm account by putting @ + your npm user name in front of the name. For example, my npm user name is ceejbot, so my scope is @ceejbot.
> http://blog.npmjs.org/post/168978377570/new-package-moniker-rules

We already have [@textlint](https://www.npmjs.com/org/textlint "@textlint") scope.

### Pros

- Make explicit

For example, [markdown-to-ast](https://github.com/textlint/textlint/tree/master/packages/markdown-to-ast "markdown-to-ast") is confused name.

### Cons

- Name chaning!

But, almost pacakges are only used in textlint.

- [textlint/packages/](https://github.com/textlint/textlint/tree/master/packages "textlint/packages/")

## Migration pacakges

| PR    | Before                     | After                                |
| :-----  | -------------------------- | ------------------------------------ |
| #473 | `txt-to-ast`               | `@textlint/text-to-ast`              |
| #460 | `markdown-to-ast`          | `@textlint/markdown-to-ast`          |
| #483 | `textlint-ast-test`        | `@textlint/ast-tester`               |
| #484 | `txt-ast-traverse`         | `@textlint/ast-traverse`             |
| #482 | `textlint-fixer-formatter` | `@textlint/fixer-formatter`          |
| #474 | `textlint-formatter`       | `@textlint/linter-formatter`         |
| #485 | `textlint-plugin-markdown` | `@textlint/textlint-plugin-markdown` |
| #486 | `textlint-plugin-text`     | `@textlint/textlint-plugin-text`     |

<details>
<summary> Migration Packages List </summary>

- [x] txt-to-ast -> `@textlint/text-to-ast` #473 
- [x] markdown-to-ast -> `@textlint/markdown-to-ast` #460 
- [x] textlint-ast-test -> `@textlint/ast-tester` #483
  - Add `er` is good
  - Align with [@textlint/ast-node-types](https://github.com/textlint/textlint/tree/master/packages/%40textlint/ast-node-types "@textlint/ast-node-types")
- [x] txt-ast-traverse -> `@textlint/ast-traverse` #484
  - Align with [@textlint/ast-node-types](https://github.com/textlint/textlint/tree/master/packages/%40textlint/ast-node-types "@textlint/ast-node-types")
[...]

PR #474, opened 2018-01-17
refactor(textlint-formatter): move to @textlint/linter-formatter

0x6b:
Ref #455.

Note: I did not change CLI name from `textlint-formatter` to `linter-formatter` for compatibility. I think `textlint-formatter` is more intuitive name for users, but if you think `linter-formatter` is better, please let me know and I will update this pull request as follows:

1. Add `linter-formatter.js` to `bin/` directory
2. Add above to `package.json#bin` as key `linter-formatter`
3. (Optional?) Add deprecate message while running `textlint-formatter` (`bin/cmd.js`)

Need major version bump as it will break CI build. Thanks.


azu:
> (Optional?) Add deprecate message while running textlint-formatter (bin/cmd.js)

No. Runtime message break exiting behavior.

Instead of it, You can deprecate old module by executing following command: 

```
npm deprecate textlint-formatter "See https://github.com/textlint/textlint/issues/455"
```

As as result, show deprecated message in installing time.

```
npm install textlint-formatter 
# show deprecated message
```

azu:
> Need major version bump as it will break CI build.

Yes and No.
These are difference name, but the API is same.
So, It does not require major update.

> ℹ️ Migration information for user
> https://github.com/textlint/textlint/issues/455

0x6b:
@azu, appreciate your comments. I missed update (and detailed!) description at #455. Run `npm deprecate` then merge the pull request. Thank you very much.

0x6b:
For the record:
```
$ npm deprecate textlint-formatter "See https://github.com/textlint/textlint/issues/455"
$ npm info --json textlint-formatter | jq .deprecated
"See https://github.com/textlint/textlint/issues/455"
```

Issue #475, opened 2018-01-17
textlint-tester: incorrect `main` and `types` declaration?


0x6b:
While I'm walking around the codebase in order to find candidate for TypeScript-ify, I found following in the `textlint-tester/package.json`:

https://github.com/textlint/textlint/blob/b769c1bbd0f60623e1c1d2cea53ad2a977179956/packages/textlint-tester/package.json#L20-L21

I think it should be:

```
 "main": "lib/textlint-tester/src/index.js", 
 "types": "lib/textlint-tester/src/index.d.ts", 
```

Any thoughts?

azu:
Opps. This is a bug.

azu:
Side note: we need to E2E test for textlint-tester.

0x6b:
@azu, thanks for prompt response and pull request!

0x6b:
Side note (2): how about sorting `package.json` using [`sort-package-json`](https://www.npmjs.com/package/sort-package-json)? It will enable us find such inconsistency more easily.

$ yarn lerna exec npx sort-package-json

If it's ok for you, I will open an issue and pull request.

azu:
Yes! 
I use `sort-package-json` in locally (just in setup)

PR #476, opened 2018-01-17
fix(tester): fix "main" and "types" field in package.json

azu:
fix https://github.com/textlint/textlint/issues/475


azu:
Travis CI stopping...

PR #486, opened 2018-01-19
refactor(plugin-text): move to @textlint

0x6b:
Finally closes #455!

Note: following line should be modified to `https://github.com/textlint/textlint/tree/master/packages/@textlint/textlint-plugin-text` after merge. I will open another pull request.

https://github.com/textlint/textlint/blob/29785dea1c91f3a6f818454a366a86033787346b/website/pages/en/index.js#L118

PR #485, opened 2018-01-19
refactor(plugin-markdown): move to @textlint

0x6b:
Ref #455.

Note: following line should be modified to `https://github.com/textlint/textlint/tree/master/packages/@textlint/textlint-plugin-markdown` after merge. I will open another pull request.

https://github.com/textlint/textlint/blob/29785dea1c91f3a6f818454a366a86033787346b/website/pages/en/index.js#L114

PR #487, opened 2018-01-20
Update links and markdown

0x6b:
A small follow-up pull request for #485 and #486. Will merge after successful CI builds as it is trivial changes.


0x6b:
@azu, thanks!

PR #491, opened 2018-01-22
feat(textlint-tester): add multiple rules and plugins test support

### LINKED ISSUE #488: textlint-tester: use with plugin

- textlint 10.1.3
- textlint-tester 4.0.6
- [textlint-plugin-html](https://github.com/textlint/textlint-plugin-html) 0.1.7

HTML

1. Have been updating dependencies for [textlint/examples/html-plugin](https://github.com/textlint/textlint/tree/master/examples/html-plugin) for housekeeping
2. Updated [`textlint-rule-sentence-length`](https://seattle.github.com/textlint-rule/textlint-rule-sentence-length) from 1.0.4 to 2.0.0
3. Found that [`sentence-splitter`](https://github.com/azu/sentence-splitter/) 3.0.6 (which `textlint-rule-sentence-length` 2.0.0 depends on) has a defect which causes infinite loop and out of memory error (azu/sentence-splitter#9)
4. @azu kindly released [3.0.7](https://github.com/azu/sentence-splitter/releases/tag/3.0.7) which solved the issue
5. Back to `textlint-rule-sentence-length`, updated `sentence-splitter` to 3.0.7
6. Added valid test case to [`textlint-rule-sentence-length.js`](https://github.com/textlint-rule/textlint-rule-sentence-length/blob/26922b6bd30f2820cd81242a77bc84cf8cecba53/test/sentence-length-test.js) as follows:

```js
{
    text: "<p>this is a test</p>",
    ext: ".html"
}
```

Test finished without error expecting textlint uses html plugin according to `ext` property described at [README](https://github.com/textlint/textlint/blob/master/packages/textlint-tester/README.md).

**What actually happened?**

Got following error:

```
  1) textlint-rule-sentence-length
       <p>this is a test</p>:
     Error: Not found available plugin for .html
      at node_modules/@textlint/kernel/src/textlint-kernel.ts:116:23
      at <anonymous>
```

Reviewed `textlint-tester.ts` and found that the tester does not initialize plugins according to `ext` property as expected.

https://github.com/textlint/textlint/blob/3d609a7f0bac001170345b93e221aa9a5faaac27/packages/textlint-tester/src/textlint-tester.ts#L79-L80

Simply adding `pluginName` and `plugin` parameters then call `setupPlugins` as follows solved the issue, but function signature looks ugly for me.

- In order to maintain compatibility, I think `pluginName` and `plugin` parameters should be optional, but
- A required parameter cannot follow an optional parameter (TS1016)

```diff
--- a/packages/textlint-tester/src/textlint-tester.ts
+++ b/packages/textlint-tester/src/textlint-tester.ts
@@ -71,12 +71,28 @@ export class TextLintTester {
         }
     }

-    testValidPattern(ruleName: string, rule: TextlintRuleCreator, valid: TesterValid) {
+    testValidPattern(ruleName: string, rule: TextlintRuleCreator, valid: TesterValid,
+        pluginName?: string,
+        plugin?: object
+    ) {
         const text = typeof valid === "object" ? valid.text : valid;
         const inputPath = typeof valid === "object" ? valid.inputPath : undefined;
         const options = (typeof valid === "object" && valid.options) || {};
         const ext = typeof valid === "object" && valid.ext !== undefined ? valid.ext : ".md";
         const textlint = new TextLintCore();
+        if (pluginName && plugin) {
+            textlint.setupPlugins({ [pluginName]: plugin }, { [pluginName]: options });
+        }
         textlint.setupRules(
             {
                 [ruleName]: rule

(omit rest for brevity)
```
Test script:
```js
import TextLintTester = require("../src/index");
const noTodo = require("textlint-rule-no-todo");
const htmlPlugin = require("textlint-plugin-html");
const tester = new TextLintTester();
tester.run(
    "no-todo", noTodo, { valid: [...], invalid: [...] },
    "textlint-plugin-html", htmlPlugin // new parameters
);
```

Possible options I can think of is:

1. Add plugin related parameters as above
2. Use [`textlint-engine-core`](https://github.com/textlint/textlint/blob/master/packages/textlint/src/engine/textlint-engine-core.ts) from `textlint-tester` (a lot of work?)
3. Leave current implementation as is, and document this behavior—can't test with plugins other than default `text` and `markdown`—as limitation
4. Any other

Appreciate your input. Thank you.

0x6b:
Closes #488. Please review. Thank you.

### DIFF
--- a/packages/textlint-tester/README.md
+++ b/packages/textlint-tester/README.md
@@ -46,7 +46,7 @@ export declare type TestConfig = {
     rules: {
         ruleId: string; // name of rule
         rule: TextlintRuleCreator; // textlint rule
-        options?: any; // optionsfor rule
+        options?: any; // options for rule
     }[];
 };
 ```
@@ -84,7 +84,7 @@ export declare type TestConfig = {
         - `{string} text`: a text to be linted
         - `{string} ext`: an extension key. Default: `.md` (Markdown)
         - `{string} inputPath`: a test text filePath that prefer to `text` property
-        - `{object} options`: options to be passed to the rule. Will be ignored if `testConfig` is specified 
+        - `{object} options`: options to be passed to the rule. Will throw assertion error if `testConfig` is specified
 
 TypeScript declaration is for valid as follows:
 
@@ -124,8 +124,8 @@ export declare type TesterValid = string | {
         - `{string} inputPath`: a test text filePath that prefer to `text` property.
         - `{string} output`: a fixed text.
         - `{string} ext`: an extension key.
-        - `{object[]} errors`: an array of error objects which should be raised againt the text.
-        - `{object} options`: options to be passed to the rule. Will be ignored if `testConfig` is specified 
+        - `{object[]} errors`: an array of error objects which should be raised against the text.
+        - `{object} options`: options to be passed to the rule. Will throw assertion error if `testConfig` is specified
 
 TypeScript declaration is as follows:
 
--- a/packages/textlint-tester/README.md
+++ b/packages/textlint-tester/README.md
@@ -1,6 +1,6 @@
 # textlint-tester
 
-[Mocha](http://mochajs.org/ "Mocha") test helper library for [textlint](https://github.com/textlint/textlint "textlint") rule.
+[Mocha](http://mochajs.org/ "Mocha") test helper library for [textlint](https://github.com/textlint/textlint "textlint") [rule](../../docs/rule.md).
 
 ## Installation
 
@@ -11,20 +11,84 @@
 1. Write tests by using `textlint-tester`
 2. Run tests by [Mocha](http://mochajs.org/ "Mocha")
 
+```sh
+$(npm bin)/mocha test/
+```
+
 ### TextLintTester
+
+`TextLintTester#run` has two signatures.
+
+- If you want to test single rule, use first method (`TextLintTester#run(ruleName, rule, {valid=[], invalid=[]})`)
+- If you want to test multiple rules and/or plugins, use second method (`TextLintTester#run(testName, testConfig, {valid=[], invalid=[]})`)
 
 #### TextLintTester#run(ruleName, rule, {valid=[], invalid=[]})
 
+- `{string} ruleName` ruleName is a name of the rule.
+- `{TextlintRuleCreator} rule` rule is textlint rule.
+
+#### TextLintTester#run(testName, testConfig, {valid=[], invalid=[]})
+
+- `{string} testName` testName is a name of the test.
+- `{TestConfig} testConfig` testConfig is the configuration object for the test.
+
+##### TestConfig object
+
+`TestConfig` is defined as follows:
+
+```typescript
+export declare type TestConfig = {
+    plugins: {
+        pluginId: string; // name of plugin
+        plugin: TextlintPluginCreator; // textlint plugin
+    }[];
+    rules: {
+        ruleId: string; // name of rule
+        rule: TextlintRuleCreator; // textlint rule
+    }[];
+};
+```
+
+`testConfig` object example:
+
+```js
+{
+    plugins: [
+        {
+            pluginId: "html",
+            plugin: htmlPlugin // = require("textlint-plugin-html")
+        }
+    ],
+    rules: [
+        {
+            ruleId: "no-todo",
+            rule: noTodoRule // = require("textlint-rule-no-todo")
+
+        }
+    ]
+}
+```
+
 ##### valid object
 
-- `{string} ruleName` ruleName is a name of the rule.
-- `{Function} rule` rule is the exported function of the rule.
 - `{string[]|object[]} valid` valid is an array of text which should be passed.
     - You can use `object` if you want to specify some options. `object` can have the following properties:
         - `{string} text`: a text to be linted
         - `{string} ext`: an extension key. Default: `.md` (Markdown)
         - `{string} inputPath`: a test text filePath that prefer to `text` property
         - `{object} options`: options to be passed to the rule
+
+TypeScript declaration is for valid as follows:
+
+```typescript
+export declare type TesterValid = string | {
+    text?: string;
+    ext?: string;
+    inputPath?: string;
+    options?: any;
+    pluginOptions?: any;
+};
+```
 
 `valid` object example:
 
@@ -45,7 +109,6 @@
 ]
 ```
 
-
 ##### invalid object
 
 - `{object[]} invalid` invalid is an array of object which should be failed.
@@ -55,6 +118,27 @@
         - `{string} output`: a fixed text.
         - `{string} ext`: an extension key.
         - `{object[]} errors`: an array of error objects which should be raised againt the text.
+
+TypeScript declaration is as follows:
+
+```typescript
+export declare type TesterInvalid = {
+    text?: string;
+    output?: string;
+    ext?: string;
+    inputPath?: string;
+    options?: any;
+    pluginOptions?: any;
+    errors: {
+        ruleId?: string;
+        index?: number;
+        line?: number;
+        column?: number;
+        message?: string;
+        [index: string]: any;
+    }[];
+};
+```
 
 `invalid` object example:
 
@@ -144,11 +228,8 @@ tester.run("no-todo", rule, {
 });
 ```
 
-Run the tests:
+See [`textlint-tester-test.ts`](./test/textlint-tester-test.ts) or [`textlint-tester-plugin.ts`](./test/textlint-tester-plugin.ts) for concrete examples.
 
-```sh
-$(npm bin)/mocha test/
-```
 
 ## Contributing
 
--- a/packages/textlint-tester/README.md
+++ b/packages/textlint-tester/README.md
@@ -38,7 +38,7 @@ $(npm bin)/mocha test/
 
 ```typescript
 export declare type TestConfig = {
-    plugins: {
+    plugins?: {
         pluginId: string; // name of plugin
         plugin: TextlintPluginCreator; // textlint plugin
         options?: any; // options for plugin
--- a/packages/textlint-tester/README.md
+++ b/packages/textlint-tester/README.md
@@ -41,10 +41,12 @@ export declare type TestConfig = {
     plugins: {
         pluginId: string; // name of plugin
         plugin: TextlintPluginCreator; // textlint plugin
+        options?: any; // options for plugin
     }[];
     rules: {
         ruleId: string; // name of rule
         rule: TextlintRuleCreator; // textlint rule
+        options?: any; // optionsfor rule
     }[];
 };
 ```
@@ -63,7 +65,13 @@ export declare type TestConfig = {
         {
             ruleId: "no-todo",
             rule: noTodoRule // = require("textlint-rule-no-todo")
-
+        },
+        {
+            ruleId: "max-number-of-lines",
+            rule: maxNumberOfLineRule, // = require("textlint-rule-max-number-of-lines")
+            options: {
+                max: 2
+            }
         }
     ]
 }
@@ -76,8 +84,7 @@ export declare type TestConfig = {
         - `{string} text`: a text to be linted
         - `{string} ext`: an extension key. Default: `.md` (Markdown)
         - `{string} inputPath`: a test text filePath that prefer to `text` property
-        - `{object} options`: options to be passed to the rule
-        - `{object} plugibOptions`: options to be passed to the plugin
+        - `{object} options`: options to be passed to the rule. Will be ignored if `testConfig` is specified 
 
 TypeScript declaration is for valid as follows:
 
@@ -87,7 +94,6 @@ export declare type TesterValid = string | {
     ext?: string;
     inputPath?: string;
     options?: any;
-    pluginOptions?: any;
 };
 ```
 
@@ -119,8 +125,7 @@ export declare type TesterValid = string | {
         - `{string} output`: a fixed text.
         - `{string} ext`: an extension key.
         - `{object[]} errors`: an array of error objects which should be raised againt the text.
-        - `{object} options`: options to be passed to the rule
-        - `{object} plugibOptions`: options to be passed to the plugin
+        - `{object} options`: options to be passed to the rule. Will be ignored if `testConfig` is specified 
 
 TypeScript declaration is as follows:
 
@@ -131,7 +136,6 @@ export declare type TesterInvalid = {
     ext?: string;
     inputPath?: string;
     options?: any;
-    pluginOptions?: any;
     errors: {
         ruleId?: string;
         index?: number;
--- a/packages/textlint-tester/README.md
+++ b/packages/textlint-tester/README.md
@@ -77,6 +77,7 @@ export declare type TestConfig = {
         - `{string} ext`: an extension key. Default: `.md` (Markdown)
         - `{string} inputPath`: a test text filePath that prefer to `text` property
         - `{object} options`: options to be passed to the rule
+        - `{object} plugibOptions`: options to be passed to the plugin
 
 TypeScript declaration is for valid as follows:
 
@@ -118,6 +119,8 @@ export declare type TesterValid = string | {
         - `{string} output`: a fixed text.
         - `{string} ext`: an extension key.
         - `{object[]} errors`: an array of error objects which should be raised againt the text.
+        - `{object} options`: options to be passed to the rule
+        - `{object} plugibOptions`: options to be passed to the plugin
 
 TypeScript declaration is as follows:
 
--- a/packages/textlint-tester/src/textlint-tester.ts
+++ b/packages/textlint-tester/src/textlint-tester.ts
@@ -258,6 +258,22 @@ export class TextLintTester {
     ) {
         if (isTestConfig(param)) {
             assertTestConfig(param);
+            if (valid) {
+                valid.forEach(validCase => {
+                    assert(
+                        !validCase.hasOwnProperty("options"),
+                        "Could not specify options property in valid object when TestConfig was passed. Use TestConfig.rules.options."
+                    );
+                });
+            }
+            if (invalid) {
+                invalid.forEach(invalidCase => {
+                    assert(
+                        !invalidCase.hasOwnProperty("options"),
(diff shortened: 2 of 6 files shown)


packages/textlint-tester/README.md
```diff
@@ -11,21 +11,85 @@
...
 ##### valid object
 
-- `{string} ruleName` ruleName is a name of the rule.
-- `{Function} rule` rule is the exported function of the rule.
 - `{string[]|object[]} valid` valid is an array of text which should be passed.
     - You can use `object` if you want to specify some options. `object` can have the following properties:
         - `{string} text`: a text to be linted
         - `{string} ext`: an extension key. Default: `.md` (Markdown)
         - `{string} inputPath`: a test text filePath that prefer to `text` property
         - `{object} options`: options to be passed to the rule
```
0x6b: I have to add `pluginOptions` here.

packages/textlint-tester/README.md
```diff
@@ -56,6 +119,27 @@
         - `{string} ext`: an extension key.
         - `{object[]} errors`: an array of error objects which should be raised againt the text.
```
0x6b: I have to add `options` and `pluginOptions` here.

azu:
Thanks for PR.
I'll look into details later.

📝 I found an unsolved issue.
When use multiple `rules` and specify `options` in valid or invalid, the result is non-deterministic behavior.
Should we need `options` syntax when set multiple rules.

```js
tester.run(
    "new-style-of-test",
    {
        rules: [
            {
                ruleId: "rule1",
                rule: rule1
            },
            {
                ruleId: "rule2",
                rule: rule2
            }
        ],
    },
    {
        valid: [
            {
                text: "<p>日本語 is Japanese.</p>",
                ext: ".html",
                options: { // <= the options affect to rule1 or rule2?
                  a: 1
                }
            }
        ]
    }
);
```

packages/textlint-tester/src/textlint-tester.ts
```diff
@@ -40,20 +40,32 @@ function assertHasFixer(ruleCreator: any, ruleName: string): any {
     throw new Error(`Not found \`fixer\` function in the ruleCreator: ${ruleName}`);
 }
 
+export type TestConfig = {
+    plugins: { pluginId: string; plugin: TextlintPluginCreator }[];
```
0x6b: I have to add `pluginOptions` and `ruleObject` here, in order to configure each plugin/rule respectively.

packages/textlint-tester/src/textlint-tester.ts
```diff
@@ -40,20 +40,32 @@ function assertHasFixer(ruleCreator: any, ruleName: string): any {
...
+    rules: { ruleId: string; rule: TextlintRuleCreator }[];
+};
+
+function isTestConfig(arg: any): arg is TestConfig {
+    return arg.plugins !== undefined;
```
azu: `rules` is required at least one.

- is `rules` array
- is `rules` length > 0
   - assert: rule item should have `ruleId` and `rule`
- is `plugins` array
   - assert: Each plugin item should have `pluginId` and `plugin`

It is better that assert item object.
It is helpful for (non-TypeScript) user.

azu: It is good that export and  test it.

Patterns:

- OK: `rules` only
- OK `rules` and `plugins`
- NG: `plugins` only
- NG: Not have property

0x6b:
@azu, thank you for comments. I have implemented new options handling for proof of concept. Please review 985d03a.

- When use `TestConfig`, ignore `options` in `valid` and `invalid`
- Add options for rule and plugin at `TestConfig`

As for assertion for non TypeScript users, I will implement it later today or tomorrow.

packages/textlint-tester/README.md
```diff
@@ -11,20 +11,91 @@
...
+- `{string} testName` testName is a name of the test.
+- `{TestConfig} testConfig` testConfig is the configuration object for the test.
+
+##### TestConfig object
+
+`TestConfig` is defined as follows:
+
+```typescript
+export declare type TestConfig = {
+    plugins: {
```
azu: `plugins?`

`plugins` is optional.

0x6b: @azu, thank you. Yes, `plugins` is optional. Reconsidering `isTestConfig` implementation...

```typescript
function isTestConfig(arg: any): arg is TestConfig {
    return arg.plugins !== undefined;
}
```

0x6b:
Hmm. New tests will fail if Node.js < 7.

packages/textlint-tester/src/textlint-tester.ts
```diff
@@ -71,47 +119,109 @@ export class TextLintTester {
...
+        if (isTestConfig(param)) {
+            const rules: any = {};
+            const rulesOptions: any = {};
+            param.rules.forEach(rule => {
+                const ruleName = rule.ruleId;
+                const ruleObject = rule.rule;
+                const ruleOptions = rule.options || {};
+                rules[ruleName] = ruleObject;
+                rulesOptions[ruleName] = ruleOptions;
+            });
```
azu: Please create a function that wrap following code:

```diff
+            const rules: any = {};
+            const rulesOptions: any = {};
+            param.rules.forEach(rule => {
+                const ruleName = rule.ruleId;
+                const ruleObject = rule.rule;
+                const ruleOptions = rule.options || {};
+                rules[ruleName] = ruleObject;
+                rulesOptions[ruleName] = ruleOptions;
+            });
```

```js
const createRuleXXX = (rules: TestConfigRule[]): XXX => { ... }
const XXX = createRuleXXX(param.rules);
textlint.setupRules(XXX.rules, XXX.rulesOptions);
```

I have not good name :(

`textlint` use [RuleCreatorSet](https://github.com/textlint/textlint/blob/master/packages/textlint/src/core/rule-creator-set.ts "RuleCreatorSet") that is similar to this.

RuleChunk, RuleDefinition, RuleGroup ...

packages/textlint-tester/src/textlint-tester.ts
```diff
@@ -71,47 +119,109 @@ export class TextLintTester {
...
```
azu: ditto.

Create `createRuleXXX` function and use it.

packages/textlint-tester/src/textlint-tester.ts
```diff
@@ -71,47 +119,109 @@ export class TextLintTester {
...
+            textlint.setupRules(rules, rulesOptions);
+            if (param.plugins !== undefined) {
```
azu: `Array.isArray(param.plugins)` is better.

0x6b:
Error#code is added since Node.js v8.

packages/textlint-tester/test/textlint-tester-invalid-testcofig-test.js
```diff
@@ -0,0 +1,150 @@
...
+                    invalid: [
+                        {
+                            text: "- [ ] string",
+                            ext: ".md",
+                            errors: [{ message: "Found TODO: '- [ ] string'", line: 1, column: 3 }]
+                        }
+                    ]
+                });
+            } catch (err) {
+                assert.equal(err.code, "ERR_ASSERTION");
```
azu: Does `err instanceof assert.AssertError` work?

azu:
> Error#code is added since Node.js v8.

Yes. [Node.js Errors — Changes you need to know about – Node.js Collection – Medium](https://medium.com/the-node-js-collection/node-js-errors-changes-you-need-to-know-about-dc8c82417f65 "Node.js Errors — Changes you need to know about – Node.js Collection – Medium")

packages/textlint-tester/src/textlint-tester.ts
```diff
@@ -71,47 +81,105 @@ export class TextLintTester {
...
+        }
         it(inputPath || text, () => {
             return testInvalid({ textlint, inputPath, text, ext, errors });
         });
         // --fix
         if (invalid.hasOwnProperty("output")) {
             it(`Fixer: ${inputPath || text}`, () => {
-                assertHasFixer(rule, ruleName);
+                if (isTestConfig(param)) {
+                    (param as TestConfig).rules.forEach(rule => {
```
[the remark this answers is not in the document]
0x6b: Oh I thought I removed such casting when implementing safe guard (77b0e95). Removed.

packages/textlint-tester/src/textlint-tester.ts
0x6b: Alright. Fixed.

packages/textlint-tester/test/textlint-tester-invalid-testcofig-test.js
0x6b: Works as expected on Node.js v{4,5,6,7,8,9} on my end. Thank you!

packages/textlint-tester/test/textlint-tester-invalid-testcofig-test.js
```diff
@@ -0,0 +1,150 @@
+// LICENSE : MIT
+"use strict";
+
+const TextLintTester = require("../src/index");
+const htmlPlugin = require("textlint-plugin-html");
+const noTodoRule = require("textlint-rule-no-todo");
+const tester = new TextLintTester();
+const assert = require("assert");
+
+const testConfigs = [
```
azu: parameterize testing is good! 👍

packages/textlint-tester/src/textlint-tester.ts
```diff
@@ -71,47 +119,109 @@ export class TextLintTester {
...
+                    const pluginOptions = plugin.options || {};
+                    textlint.setupPlugins(
+                        {
+                            [pluginName]: pluginObject
+                        },
+                        {
+                            [pluginName]: pluginOptions
+                        }
+                    );
+                });
```
azu: we can apply same approch to setupPlugins.
Create a wrap `createPluginXXX` and use it.

packages/textlint-tester/src/textlint-tester.ts
```diff
@@ -71,47 +119,109 @@ export class TextLintTester {
...
+                rules[ruleName] = ruleObject;
+                rulesOptions[ruleName] = ruleOptions;
+            });
+            textlint.setupRules(rules, rulesOptions);
+            if (param.plugins !== undefined) {
+                param.plugins.forEach(plugin => {
+                    const pluginName = plugin.pluginId;
+                    const pluginObject = plugin.plugin;
+                    const pluginOptions = plugin.options || {};
+                    textlint.setupPlugins(
```
azu: ditto

packages/textlint-tester/src/textlint-tester.ts
```diff
@@ -40,6 +40,53 @@ function assertHasFixer(ruleCreator: any, ruleName: string): any {
...
 
+function assertTestConfig(testConfig: TestConfig): any {
+    assert.notEqual(testConfig, null, "TestConfig is null");
+    assert.notEqual(
+        Object.keys(testConfig).length === 0 && testConfig.constructor === Object,
+        true,
+        "TestConfig is empty"
+    );
+    assert(Array.isArray(testConfig.rules), "TestConfig.rules should be an array");
+    assert.notEqual(testConfig.rules.length, 0, "TestConfig.rules should have at least one rule");
```
azu: Prefer to use [assert.notStrictEqual](https://nodejs.org/api/assert.html#assert_assert_notstrictequal_actual_expected_message "assert.notStrictEqual").
Or jsut use `assert.ok(testConfig.rules.length > 0)`

packages/textlint-tester/src/textlint-tester.ts
```diff
@@ -71,47 +119,109 @@ export class TextLintTester {
...
         const ext = typeof valid === "object" && valid.ext !== undefined ? valid.ext : ".md";
         const textlint = new TextLintCore();
-        textlint.setupRules(
-            {
-                [ruleName]: rule
-            },
-            {
-                [ruleName]: options
+        if (isTestConfig(param)) {
+            const rules: any = {};
```
azu: Can use `{ [index: string]: TextlintRuleCreator; }` instead of any.

- [Index Signatures · TypeScript Deep Dive](https://basarat.gitbooks.io/typescript/docs/types/index-signatures.html "Index Signatures · TypeScript Deep Dive")

packages/textlint-tester/src/textlint-tester.ts
0x6b: Created `createTestRuleSet` function.

packages/textlint-tester/src/textlint-tester.ts
0x6b: Created `createTestRuleSet` function. Naming is difficult...

packages/textlint-tester/src/textlint-tester.ts
0x6b: Fixed with `assert.ok(testConfig.rules.length > 0)` for simplicity.

packages/textlint-tester/src/textlint-tester.ts
0x6b: An index signatures! Fixed.

packages/textlint-tester/src/textlint-tester.ts
0x6b: Created `createTestPluginSet` function.

packages/textlint-tester/src/textlint-tester.ts
0x6b: As above created `createTestPluginSet` function.

0x6b:
Finally passed CI. @azu please give me another round of review. Thank you very much.

packages/textlint-tester/README.md
```diff
@@ -11,20 +11,91 @@
...
 
-- `{string} ruleName` ruleName is a name of the rule.
-- `{Function} rule` rule is the exported function of the rule.
 - `{string[]|object[]} valid` valid is an array of text which should be passed.
     - You can use `object` if you want to specify some options. `object` can have the following properties:
         - `{string} text`: a text to be linted
         - `{string} ext`: an extension key. Default: `.md` (Markdown)
         - `{string} inputPath`: a test text filePath that prefer to `text` property
-        - `{object} options`: options to be passed to the rule
+        - `{object} options`: options to be passed to the rule. Will be ignored if `testConfig` is specified
```
azu: > Will be ignored if `testConfig` is specified 

It looks that "if specified `testConfig` property in valid object, ignore `options` property".

1. if specified `testConfig` property in `valid` object, ignore `options` property in `valid` object.
2. if specified `testConfig` property in `run` paramerter, ignore `options` property  in `valid` object

Which is correct answer?
I think that both is correct, but we need to write clearly.


`1` means that you can override `run`'s `testConfig` parameter by `valid`'s `testConfig` property.
`2` means that you can not override `run`'s `testConfig` parameter in each test case.

`2` have more restriction than `1`.
which is better? (It is case by case.)

packages/textlint-tester/test/textlint-tester-plugin.ts
```diff
@@ -0,0 +1,106 @@
...
+"use strict";
+
+import TextLintTester = require("../src/index");
+
+const htmlPlugin = require("textlint-plugin-html");
+const noTodoRule = require("textlint-rule-no-todo");
+const maxNumberOfLineRule = require("textlint-rule-max-number-of-lines");
+const tester = new TextLintTester();
+
+tester.run(
```
azu: lack of valid case

- Pass a TestConfig that has multiple `rules` only
- https://github.com/textlint/textlint/pull/491/files#r163427444

packages/textlint-tester/README.md
azu: And, We should add an assertion for invalid case after this.

azu: > Will be ignored if `testConfig` is specified 

1. if specified `testConfig` property in `valid` object, ignore `options` property".
2. if specified `testConfig` parameter in `run` method, ignore `options` property".

Which is correct?
I think that both is correct, but we need to write clearly.

`1` means that you can override `run`'s `testConfig` parameter by `valid`'s `testConfig` property.
`2` means that you can not override `run`'s `testConfig` parameter in each test case.

`2` have more restriction than  `1`.
Which is better? It is case by case, I think.

azu: We need to add an assertion for invalid pattern after this.

packages/textlint-tester/test/textlint-tester-invalid-testcofig-test.js
```diff
@@ -0,0 +1,150 @@
...
+                    ruleId: "no-todo",
+                    rule: noTodoRule
+                }
+            ]
+        },
+        expectedErrorMessage: "plugin property not found"
+    }
+];
+
+describe("new-style-of-test: invalid testConfig", () => {
```
azu: lack of invalid case

- https://github.com/textlint/textlint/pull/491#discussion_r163429604 assert invalid `options` pattern.

packages/textlint-tester/src/textlint-tester.ts
```diff
@@ -64,54 +112,116 @@ export type TesterInvalid = {
...
         const ext = invalid.ext !== undefined ? invalid.ext : ".md";
         const textlint = new TextLintCore();
-        textlint.setupRules(
-            {
-                [ruleName]: rule
-            },
-            {
-                [ruleName]: options
+        if (isTestConfig(param)) {
+            const testRuleSet = createTestRuleSet(param.rules);
```
azu: Assert: `options` property is invalid, or ignored.

If `options` is passed, to throw an error is helpful.

packages/textlint-tester/test/textlint-tester-plugin.ts
0x6b: I see. Will add such case. (I tried to salvage notification from my inbox but no avail...sorry for that)

packages/textlint-tester/README.md
0x6b: @azu, thanks a ton for your continuous review.

What I meant is 2. It should be good enough for now. Overriding gives test case writer more flexibility but writing configuration for multiple rules could be complex. I think we can implement option 1 if we found such use case. I will work on 2 first. As you said need more clear explanation so will update README as well.

azu: Yes. I agree with you.
`2` is enough as a first step.

0x6b:
@azu, could you make some time to review additional commits which might address your review. Thank you.

0x6b:
@azu, much appreciate your help!

PR #501, opened 2018-03-15
refactor(ast-tester): Convert to TypeScript

azu:


azu:
fmm, build failed on windows.

PR #527, opened 2018-06-04
Freeze Context

### LINKED ISSUE #508: textlint: Freeze Context
Similar with #489 

Currently, `RuleContext` is not freeze.
It can be modified by rule user.

We want to prevent modification shread object like RuleContext.

## Task

- [ ] Object.freeze RuleContext

Leko:
Closes #508

### DIFF
--- a/packages/@textlint/kernel/src/task/fixer-task.ts
+++ b/packages/@textlint/kernel/src/task/fixer-task.ts
@@ -63,6 +63,7 @@ export default class TextLintCoreTask extends CoreTask {
             report,
             configBaseDir: this.configBaseDir
         });
+        Object.freeze(ruleContext);
         const ruleModule = getFixer(this.fixerRule.rule);
         this.tryToAddListenRule(ruleModule, ruleContext, this.fixerRule.options);
         // setup "filters" field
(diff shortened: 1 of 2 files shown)


packages/@textlint/kernel/src/task/fixer-task.ts
```diff
@@ -63,6 +63,7 @@ export default class TextLintCoreTask extends CoreTask {
             report,
             configBaseDir: this.configBaseDir
         });
+        Object.freeze(ruleContext);
```
azu: You can create a factory function like `createRuleContext`.

```ts
export const createFreezedRuleContext = (args: RuleContextArgs) => {
   return Object.freeze(new RuleContext(arg));
}
// FIXME: default export => named export
export default class RuleContext implements BaseRuleContext {
}
```

This pattern is type-safe and I like it.

```ts
const obj = { key: "str" }
const freezedObject = Object.freeze(obj);
freezedObject.key = "test"; // freezedObject is read only
```

- https://agentcooper.github.io/typescript-play/?noImplicitAny=false&target=2#code/MYewdgzgLgBCBGArGBeGBvGBrApgTwC4YAiaAJ2JgF8AoUSWAMzJxwC8cATAeSR2FhpeiflAB0zVhwAUCRAEoA3DUnsuw0WNx5UJKDmjFFQA

Leko: I'm done. How do you feel about my changes?

PR #538, opened 2018-07-13
refactor(kernel): merge textlintrc-descriptor to kernel

azu:
- Merge  textlintrc-descriptor to kernel
  - It prevent a circular reference between descriptor and kernel
  - type only module is not good idea. because, the type depended on implementation
- `@textlint/kernel` exports  `TextlintkernelDescriptor` classes


azu:
kernel should accept ` TextlintkernelDescriptor` object directly If possible.
It will improve performance to prevent creating duplicated object.

Issue #544, opened 2018-08-04
yarn prettier does not work on Windows


otariidae:
**What did you do? Please include the actual source code causing the issue.**
When I was working on developing, I ran `yarn prettier`.

**What did you expect to happen?**
I expected `yarn prettier` worked well,

**What actually happened? Please include the actual, raw output from textlint.**

`yarn prettier` threw an error.

```
~\textlint>yarn prettier
yarn run v1.9.2
$ prettier --write 'packages/**/*.{js,jsx,ts,tsx,css}'
[error] No matching files. Patterns tried: 'packages/**/*.{js,jsx,ts,tsx,css}' !**/node_modules/** !./node_modules/**
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I tried some patterns in package.json to fix it: 
`    "prettier": "prettier --write 'packages/**/*.{js,jsx,ts,tsx,css}'",` : Single quotes (current version). This does not work.
`    "prettier": "prettier --write packages/**/*.{js,jsx,ts,tsx,css}",` : No quotes. This works.
`    "prettier": "prettier --write \"packages/**/*.{js,jsx,ts,tsx,css}\"",` : Escaped double quotes. This also works.

tried in Command Prompt on latest version of Windows10. I don't know how these patterns work on Mac and Linux.

azu:
> Escaped double quotes.

Thanks. Escaped double quotes is correct.

https://github.com/eslint/eslint/issues/5796#issuecomment-206765069

Issue #453, opened 2018-01-10
Stability of textlint-rule-no-dead-link 


azu:
textlint-rule-no-dead-link is introduced by #442 
This rule includes false-positive in principle.
Because, this rule depended on server status.

We have some options for resolving this issue.

- Make `no-dead-link`'s severity  `"warning"`
  - Simple option
- Enhance textlint-rule-no-dead-link it-self
  - Improve timeout error handling
- Report lint result by comment instead of exit status
  - It can prevent to fail CI. Instead of, a bot comment the lint results.
  - [Danger JS](http://danger.systems/js/ "Danger JS") and [reviewdog](https://github.com/haya14busa/reviewdog "reviewdog") help this option

0x6b:
> Enhance textlint-rule-no-dead-link it-self

Just an idea. 40x ([client error](https://tools.ietf.org/html/rfc7231#section-6.5)) should be fixed immediately while 50x could be ignored because (some of) it indicates intermittent error on server side. Specify severity by its HTTP status code might reduce false–positive?

Leko:
I agree with @0x6b idea to reduce false–positive.

- 50x: Not error
- 40x: Error
- (301: warning?)

azu:
[textlint-rule-no-dead-link](https://github.com/textlint-rule/textlint-rule-no-dead-link) aim to find dead link.
5xx response is a part of dead link. The reason come from server.

For example, If a server response "Internal Error 500" for a long time, this link may be dead.

If the 5xx is **temporary** error, the rule can ignore it.
But, the rule(client side) can not know the true.

> (301: warning?)

This is a fixable error.
https://github.com/textlint-rule/textlint-rule-no-dead-link#obsolete-link-detection

Leko:
Hmm...
Then, I think it's better: `Report the results with comments instead of the end state`.
If error mute as warning, it prevent false-positive but I think it is dead letter.

It is to take time on each PR, so probably we don't look these warnings.
It will increase the build time meanlessly.

melg8:
Encountered same issue with some runs triggering:
``` bash
  62:16  error  https://github.com/textlint/textlint is dead. (429 too many requests)            no-dead-link
```
and some times - clean run.

azu:
I've written a recipe: [textlint + SARIF output + code scanning](https://github.com/textlint-rule/textlint-rule-no-dead-link#textlint--sarif-output--code-scanning)

PR #545, opened 2018-08-04
fix(prettier): use escaped double quotes instead of single quotes

otariidae:
Thank you for [your quick response](https://github.com/textlint/textlint/issues/544#issuecomment-410427813).

I changed package.json. Now prettier uses escaped double quotes.


azu:
Note: CI was failed by temporary error #453 
Your PR is no problem.
> no-dead-link: https://git-scm.com/book/gr/v2/Customizing-Git-Git-Hooks is dead. (500 Internal Server Error)

PR #578, opened 2019-01-24
Fix typo in link to rules page

mockdeep:


azu:
Thanks for PR.

Someone modify wiki's url and break the url.
I've reverted the wiki's page. 
https://github.com/textlint/textlint/wiki/Collection-of-textlint-rule

Thanks

PR #594, opened 2019-03-27
docs(use-as-modules): remove unnecessary char

suzuki:
## About

I think there is no need for this character

screenshot


azu:
Thanks!

PR #608, opened 2019-07-04
chore(deps): update diff library

azu:
It fix CVE alert.

3 https://github.com/kpdecker/jsdiff/blob/master/release-notes.md#v300---august-23rd-2016


azu:
We will release next patch udpate as soon as.

PR #644, opened 2019-11-02
Add configuration support for "textlint" field in package.json

trevordmiller:
Closes https://github.com/textlint/textlint/issues/643

PR #645, opened 2019-11-03
test(textlint): add text for package.json's "textlint"

azu:
Add tests and example/config-in-package-json for #644

### DIFF
--- a/packages/textlint/test/config/config-as-example-test.js
+++ b/packages/textlint/test/config/config-as-example-test.js
@@ -7,16 +7,20 @@ import * as path from "path";
 import { Config } from "../../src/config/config";
 /* load config from "./config/" and match expected result */
 describe("config-as-example", function() {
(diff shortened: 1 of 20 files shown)


packages/textlint/test/config/config-as-example-test.js
```diff
@@ -7,16 +7,20 @@ import * as path from "path";
 import { Config } from "../../src/config/config";
 /* load config from "./config/" and match expected result */
 describe("config-as-example", function() {
-    const configList = glob.sync(path.join(__dirname, "/config-fixtures/**/.textlintrc"));
+    const configList = glob.sync(path.join(__dirname, "/config-fixtures/*/{.textlintrc,package.json}"));
```
azu: Reconnize `package.json`

azu:
failed is expected. It is includes non-pushed dir.

PR #650, opened 2020-01-06
feat(types): allow plugin's `preProcess` return `{ text, ast }`

### LINKED ISSUE #649: Feature Suggestions: textlint-plugin supports binary format
v11.5.0

etc(binary file type)

I’m thinking of create [textlint-plugin for pptx](https://github.com/heriet/textlint-plugin-pptx). The pptx is binary file format. So, textlint can't load decoded `text` from pptx file.

A textlint-plugin for binary format will return correct TxtAST, but textlint kernel's [getSource](https://github.com/textlint/textlint/blob/textlint%4011.5.0/packages/@textlint/kernel/src/context/TextlintSourceCodeImpl.ts#L87) generates source from file `text`. That has lead to textlint-plugin for binary format is not work.

I suggest that [getSource](https://github.com/textlint/textlint/blob/textlint%4011.5.0/packages/@textlint/kernel/src/context/TextlintSourceCodeImpl.ts#L87) generates source from `node.raw`. If `beforeCount` or `afterCount` is inputed, getSource generetes source from `parent node.raw`.

textlint-plugin-pptx returns following AST.

```
{
  type: 'Document',
  raw: 'hoge'
  ...
}
```

and prh rule.

```
rules:
  - expected: fuga
    patterns:
      - hoge
```

Expected

```
$ textlint example.pptx
example.pptx
  1:1   ✓ error  hoge => fuga  prh
✖ 1 problems (1 errors, 0 warnings)
```

textlint is not work because getSource returns from `text` (encoded binary) and textlint rules using getSource.

```
$ textlint example.pptx
```

**Note**

- If textlint don't want to support textlint-plugin for binary format file, we can also choose alternateve way that is convert binary format to markdown before textlint.
- `--fix` will not work textlint-plugin for binary format. but I think it is small problem.
- Alternateve way for support binary format is extending textlint-plugin Processor. For example,  Processor preProcess returns decoded text for source. But I think it may be difficult.

azu:
This PR allow plugin creator to create a plugin for binary file.

Currently, plugin just return an AST object for the target file.
This changes make plugin return a Text and an AST object for the target file.

fix #649


packages/textlint/test/pluguins/fixtures/example-plugin.js
```diff
@@ -14,6 +14,8 @@ export class ExampleProcessor {
                 return {
                     type: "Document",
                     children: [],
+                    value: "",
```
azu: `isTxtAST` check that dummy properties.

Issue #585, opened 2019-02-08
Reduce unnecessary dependencies(Node 6>=)


azu:
textlint support Node.js 6 >=.( We plan to drop Node.js 6 in addition #600 )
So, we can use native function without polyfill.
For example, we can use `Object.assign` instead of `object-assign` module.

## Tasks

- [x] use `Object.assign` instead of `object-assign` module. #618 
  - or, just use `{ ...object }` spread syntax
- [x] use `Promise` instead of `bluebird`
  - need to check performance regiression, but it is minor problem.

ia17011:
I'm looking around the issues on this project to find the task I can work on. 
I think this task is already closed by above #618 #672.

azu:
Thanks!
We can close it.

PR #660, opened 2020-02-05
refactor(textlint/kernel): use promise instead of bluebird

numb86:
This PR is for #585 .


azu:
@numb86 Thanks! Good!

Issue #669, opened 2020-04-15
Convert to TypeScript again!


azu:
We have conveted core JavaScript packages to TypeScript packages in https://github.com/textlint/textlint/issues/248

But, some packages still use JavaScript.
We want to complete migration from JavaScript to TypeScript.

## Packages

- [x] [textlint/packages/@textlint/markdown-to-ast at master · textlint/textlint](https://github.com/textlint/textlint/tree/master/packages/%40textlint/markdown-to-ast) by @azu #671 
- [x]  [textlint/packages/@textlint/text-to-ast at master · textlint/textlint](https://github.com/textlint/textlint/tree/master/packages/%40textlint/text-to-ast) by @anoriqq #673 
- [x] [textlint/packages/@textlint/textlint-plugin-text at master · textlint/textlint](https://github.com/textlint/textlint/tree/master/packages/%40textlint/textlint-plugin-text) by @Kesin11 #674 
- [x]  [textlint/packages/@textlint/textlint-plugin-markdown at master · textlint/textlint](https://github.com/textlint/textlint/tree/master/packages/%40textlint/textlint-plugin-markdown) by @anoriqq #675 

At first, convert `src/*.js` to `src/*.ts`. Test code use `allowJs` config like https://github.com/textlint/textlint/blob/master/packages/textlint/test/tsconfig.json

Welcome to contribute this migration!

**Blocker**

We are migrating hacky [module resolution](https://github.com/textlint/textlint/issues/440) to Project references #668 

- [x] Use Project References #668

azu:
Add project references usage to https://github.com/textlint/textlint/blob/master/docs/CONTRIBUTING.md#note-typescripts-project-references

azu:
I touch @textlint/markdown-to-ast

anoriqq:
I'm willing to try out this issue!
I'll pick @textlint/text-to-ast up.

azu:
@anoriqq Yeah! Thanks!

I've converted pacakge by following steps in #671 

1. install `typescript ts-node ts-node-test-register` and some `@types/*` pacakges as devDependendencies *1
  - Ref: https://github.com/textlint/textlint/blob/b879a09fb0ba13775cc3d4c83f0b8054e2789f81/packages/%40textlint/markdown-to-ast/package.json#L44-L56
2. Copy `tsconfig.json` to `<pkg>/tsconfig.json`
  - This project use extended `tsconfig.base.json`
[...]

PR #673, opened 2020-04-17
refactor(text-to-ast): convert to typescript 

anoriqq:
- Convert `@textlint/text-to-ast` to TypeScript code

Ref. #669


packages/@textlint/text-to-ast/src/plaintext-parser.ts
```diff
@@ -72,7 +73,7 @@ function createBRNode(lineNumber, startIndex) {
  * @param {[TxtNode]} nodes
  * @returns {TxtNode} Paragraph node
  */
-function createParagraph(nodes) {
+function createParagraph(nodes: [TxtNode]): TxtNode {
```
azu, line 76:
```suggestion
function createParagraph(nodes: TxtNode[]): TxtNode {
```

`TxtNode[]` is Array of TxtNode.
`[TxtNode]` is an item in a array.

packages/@textlint/text-to-ast/src/plaintext-syntax.ts
```diff
@@ -1,11 +1,11 @@
 // LICENSE : MIT
 "use strict";
-const { ASTNodeTypes } = require("@textlint/ast-node-types");
-const exports = {
+import { ASTNodeTypes } from "@textlint/ast-node-types";
+
+export default {
```
azu, line 5: Can you use named exports as possible?

```suggestion
export const Syntax = {
```

Also need to change to `import { Syntax }`

azu: https://basarat.gitbook.io/typescript/main-1/defaultisbad

packages/@textlint/text-to-ast/src/plaintext-parser.ts
anoriqq: I also fixed jsdoc.

azu:
Thanks!

PR #674, opened 2020-04-18
refactor(textlint-plugin-text): Convert to TypScript

Kesin11:
Convert `@textlint/textlint-plugin-text` to TypeScript.

@azu I have some questions. I commented on each code.

Ref. #669


packages/@textlint/textlint-plugin-text/src/index.ts
```diff
@@ -1,6 +1,6 @@
 // LICENSE : MIT
 "use strict";
 import { TextProcessor } from "./TextProcessor";
-module.exports = {
+export default {
```
Kesin11, line 4: I use default exports for respect to test code.
If you want to use named exports, I'll fix it.

packages/@textlint/textlint-plugin-text/test/TextProcessor-test.ts
```diff
@@ -61,7 +61,7 @@ describe("TextProcessor", function () {
     });
     context("when target is text", function () {
         it("should report error", function () {
-            return lintText("TODO: this is todo", ".txt").then((results) => {
```
Kesin11, line 64: I don't know why this code works.
However, this argument looks obviously wrong.

packages/@textlint/textlint-plugin-text/src/TextProcessor.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    availableExtensions() {
+        return [".txt", ".text"].concat(this.extensions);
+    }
+
+    processor(_ext: string) {
+        return {
+            preProcess(text: string, _filePath?: string): TxtParentNode | { text: string; ast: TxtParentNode } {
+                // FIXME: remove any
+                return parse(text) as any;
```
Kesin11, line 25: https://github.com/textlint/textlint/blob/953eb036d4a4b6a3249a2eaaa9af27e4da7b69ea/packages/%40textlint/types/src/Plugin/TextlintPluginModule.ts#L45
TextlintPluginProcessor.preProcess need return `TxtParentNode | { text: string; ast: TxtParentNode }` but `parse()` from "@textlint/ast-node-types" return `TxtNode` type. 

So I couldn't resolve this mismatch of types. Could you give me some advice on how to resolve it?

packages/@textlint/textlint-plugin-text/src/index.ts
azu: >  If you want to use named exports, I'll fix it.

Yes, we prefer to use named exports like https://github.com/textlint/textlint/pull/673#discussion_r410262861

packages/@textlint/textlint-plugin-text/src/TextProcessor.ts
azu: Actually, https://github.com/textlint/textlint/blob/953eb036d4a4b6a3249a2eaaa9af27e4da7b69ea/packages/%40textlint/text-to-ast/src/plaintext-parser.ts#L128 `parse` return a `TxtParentNode`.
It means that the returned AST has `children` property.

packages/@textlint/textlint-plugin-text/src/index.ts
Kesin11: I fixed it on 31e0f92

packages/@textlint/textlint-plugin-text/src/TextProcessor.ts
azu: Probably, The truth is that `preProcess` return type is wrong.
textlint's core also support `TxtNode`(not have `children` property) as `preProcess`'s return value. possibly.

We have two options.

1. fix `preProcess` return type
2. fix parse return type

packages/@textlint/textlint-plugin-text/src/index.ts
```diff
@@ -1,6 +1,6 @@
 // LICENSE : MIT
 "use strict";
 import { TextProcessor } from "./TextProcessor";
-module.exports = {
+export const TextPlugin = {
```
azu, line 4: The textlint plugin need to export `{ Processor }` object.

https://textlint.github.io/docs/plugin.html

Kesin11: I found `textlint` test was failed for another reason.  It comes from wrong way of importing in "textlint-core.ts".

I fixed it at these commits.
32bcad1
d8f3432

packages/@textlint/textlint-plugin-text/src/TextProcessor.ts
Kesin11: Unfortunately, I don't know the entire of textlint implementation. So I'll follow your decision.

azu: Can you fix `preProcess` return type?

```diff
diff --git a/packages/@textlint/types/src/Plugin/TextlintPluginModule.ts b/packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
index 923dc1c9..436ed944 100644
--- a/packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
+++ b/packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
@@ -1,5 +1,5 @@
 // Plugin
-import { TxtParentNode } from "@textlint/ast-node-types";
+import { TxtNode } from "@textlint/ast-node-types";
 
 /**
  * textlint plugin option values is object or boolean.
@@ -42,7 +42,7 @@ export declare class TextlintPluginProcessor {
          * @param text
          * @param filePath
          */
-        preProcess(text: string, filePath?: string): TxtParentNode | { text: string; ast: TxtParentNode };
+        preProcess(text: string, filePath?: string): TxtNode | { text: string; ast: TxtNode };
         postProcess(messages: Array<any>, filePath?: string): { messages: Array<any>; filePath: string };
     };
 }

```

packages/@textlint/textlint-plugin-text/src/index.ts
```diff
@@ -1,6 +1,6 @@
 // LICENSE : MIT
 "use strict";
 import { TextProcessor } from "./TextProcessor";
-module.exports = {
+export const textPlugin = {
```
azu, line 4: Use default export. It is public API.
We can not change this because it does breaking change.

We prefer to use named import/export internally, But Public API is not that.

azu:
```suggestion
export default {
```

azu: textlint plugin need to export as default.
It is convetional rule.

https://textlint.github.io/docs/plugin.html

packages/@textlint/textlint-plugin-text/src/TextProcessor.ts
Kesin11: I fixed at 7880037
Is this BREAKING CHANGES? I just filled "chore" in the commit message, but if it's wrong, fix it.

packages/@textlint/textlint-plugin-text/src/index.ts
Kesin11: Oh, I'm sorry. I try to revert commit.

packages/@textlint/textlint-plugin-text/src/TextProcessor.ts
azu: It is not breaking change.
If A node which is suitable for TxtParentNode, the node is suitable for TxtNode.

Also typing change is minor in textlint
https://github.com/textlint/textlint#semantic-versioning-policy

Kesin11: All right. Should I fix this commit message to "feat(types)" following conventional commit rule and push force?

packages/@textlint/textlint-plugin-text/src/index.ts
Kesin11: textlint (textlint/packages/textlint) test:unit failed when revert to default export, so I fix `require` to `import` in textlint-core.ts
e26a747

Kesin11: https://github.com/textlint/textlint/blob/953eb036d4a4b6a3249a2eaaa9af27e4da7b69ea/packages/%40textlint/kernel/src/descriptor/TextlintPluginDescriptor.ts#L33-L36

I also found this error message may need need to correct using `export default`.

packages/@textlint/textlint-plugin-text/src/TextProcessor.ts
azu: No problem.

azu:
@Kesin11 Thanks!
I've merged it.

Kesin11:
@azu Thank you so much for your kind review 😄

PR #675, opened 2020-04-20
refactor(textlint-plugin-markdown): convert to typescript

anoriqq:
- Convert `@textlint/textlint-plugin-markdown` to TypeScript

Ref. #669


packages/@textlint/textlint-plugin-markdown/src/index.ts
```diff
@@ -2,6 +2,6 @@
 "use strict";
 import { MarkdownProcessor } from "./MarkdownProcessor";
 
-module.exports = {
+export default {
```
azu, line 5:
```
Error: Plugin should have Processor property.
module.exports = {
  Processor: class Processor{ ... }
}
```

Ah, `textlint` use `require`.
https://github.com/textlint/textlint/blob/31c3148544621ae08f068eadacacfe8340922924/packages/textlint/src/textlint-core.ts#L30
We should change `require` to `import`.

Note: plugin loader support `module.exports` and `export` statement.

https://github.com/textlint/textlint/blob/31c3148544621ae08f068eadacacfe8340922924/packages/textlint/src/config/plugin-loader.ts#L81

anoriqq: Thank you! I fixed it on a1646b95aad18baeb3209dc4caef04404c7634f4.

azu:
Thanks!

PR #679, opened 2020-05-14
docs(textlint): fix broken link

lowr:
Fixed some broken links, which seem to be introduced during migration to typescript.

I also found some files referred in the `README.md` no longer exist (e.g. `core/` directory and `rule-manager-js`), but left them untouched as I wasn't sure whether they'd been renamed or just removed.


azu:
Thanks!

Issue #600, opened 2019-04-30
Drop Node.js 10.x


azu:
[nodejs/Release: Node.js Foundation Release Working Group](https://github.com/nodejs/Release)

Node.js 6.x is EOL at 2019-04-30

## Tasks

- [x] CI #598 
- [ ] Update `engines` field 
- [ ] major update

azu:
Node 10 will be dropped in textlint 12

PR #680, opened 2020-06-06
Drop string.prototype.padstart

takenspc:
Node.js 8+ supports padStart so that we don't need the  polyfill.

See https://github.com/textlint/textlint/issues/600


azu:
Thanks!

Issue #672, opened 2020-04-16
Migrate mocha.opts to mocharc.json


azu:
mocha@7 deprecated `mocha.opts`.
We need to migrate mocha.opts to mocharc.json.

I've created migration script https://github.com/azu/mocha-migrate

##  Task

- [ ] Upgrade to Mocha@7 Probably already done?
- [ ] Migrate mocha.opts to mocharc.json

Useful oneliner:

```
./node_modules/.bin/lerna exec --no-bail npx @azu/mocha-migrate migrate-opts -- -file test/mocha.opts -type json -rm
```

## Related

- [Migrate mocha.opts to .mocharc.(json|js|yml|yaml) · Issue #125 · secretlint/secretlint](https://github.com/secretlint/secretlint/issues/125)

numb86:
I forgot to write `resolve #672` when I made the PR, but I think this issue can be closed.
I think #585 can close for the same reason too.

PR #682, opened 2020-06-09
test: migrate mocha.opts to .mocharc.json

numb86:
Migrate `mocha.opts` to `.mocharc.json`.

- `gulp-textlint` is used `6.x` mocha, So changed it to the same version as other packages.
- I seem that `packages/@textlint/kernel/test/descriptor/mocha.opts` is an unnecessary file. So removed it.

ref #672

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -1,0 +1,5 @@
+{
+  "require": [
+    "babel-register"
+  ]
+}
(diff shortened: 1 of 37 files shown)


azu:
> I seem that `packages/@textlint/kernel/test/descriptor/mocha.opts` is an unnecessary file. So removed it.

Yes. It histtorical reason. It is not nessary now.

PR #684, opened 2020-07-10
Add TypeScript support for ESLint

zweimach:
Add TypeScript support for ESLint with the awesome TypeScript ESLint project.
Beware that this PR uses type-aware lint rules and triggers tons of errors.
But fortunately, you can turn off whatever type-aware rules just like ordinary ESLint rules.


azu:
@zweimach Thanks for PR!
I've created an issue about ESLint for TypeScript.
https://github.com/textlint/textlint/issues/685

We need to multiple tasks until enable ESLint for TypeScript in CI.
It is good that these tasks are implemented by separated pull request, I think.
What do you think about it?

azu:
I've merged it.
If you can improve it continuously, please comments #685

PR #715, opened 2020-11-03
fix(readme): missing https in image path for comparison between PL and NL

davidsonfellipe:
- Missing https in image path for comparison between PL and NL

### Screenshot
Screen Shot 2020-11-03 at 9 51 52 AM


azu:
Thanks

Issue #731, opened 2020-12-22
kernel: textlint doesn't apply adjacent and not overlapping fixes


takenspc:
**What version of textlint are you using?**

11.7.6

**What file type (Markdown, plain text, etc.) are you using?**

Markdown

**What did you do? Please include the actual source code causing the issue.**

Currently textlint doesn't apply some fixes if end of one fix's range is same as start of other fixes' range.

https://github.com/textlint/textlint/blob/375a08ef964062bd992aa5a5c620639bb0c5bbb7/packages/%40textlint/kernel/test/source-code-fixer/source-code-fixer-test.js#L235-L241

This prevents textlint to apply adjacent fixes.

```js
const sourceCode = new TextlintSourceCodeImpl({
    text: "abc",
    ast: parse("abc"),
    ext: ".md"
});

// Capitalize text ("abc" to "ABC")
const result = SourceCodeFixer.applyFixes(sourceCode, [
    {
        message: "This will be applied",
        fix: {
            range: [2, 3],
            text: "C"
        }
    },
    {
        message: "This will NOT be applied",
        fix: {
            range: [1, 2],
            text: "B"
        }
    },
    {
        message: "This will be applied",
        fix: {
            range: [0, 1],
            text: "A"
        }
    }
]);
console.log(result.output); // => I got 'AbC' (my expectation is 'ABC')
```

As we can see, ranges don't overlap each other when the end of one fix's range is same as start of other fixes' range. Because end of range is exclusive. If a range is `[0, 1]`, the fix will modify `text[0]` to **before** `text[1]` so that `text[1]` will not be modified.

I would like to make textlint apply all fixes in this conditions.

azu:
fixed in https://github.com/textlint/textlint/releases/tag/textlint%4011.7.7 #732

PR #732, opened 2020-12-22
fix: make textlint apply adjacent fixes

takenspc:
Please refer to https://github.com/textlint/textlint/issues/731

The ranges don't overlap when the end of one fix's range is same as start of other fixes' range. 

This PR makes textlint apply all fixes in this condition.


azu:
Thanks!

PR #734, opened 2021-01-06
docs(website): add Soracom as one of users

0x6b:
Please add [Soracom](https://soracom.jp) as one of users. We are using textlint internally to improve our text communication, integrating this with [Zendesk](https://www.zendesk.com/) through AWS Lambda. You can see small article on our efforts at [コールセンタージャパン 2020 年 12 月号](https://callcenter-japan.com/magazine/4989.html).


azu:
Thanks!

PR #736, opened 2021-01-21
feat(source-code-fixer): add `@textlint/source-code-fixer`

### LINKED ISSUE #735: kernel: export source-code-fixer.ts
I'm creating https://github.com/textlint/editor.
It require `applyFixableMesage({ text, messsages })` like API.

Currently, @textlint/kernel proveides two APIs.

- `lintText(text: string, options: TextlintKernelOptions): Promise<TextlintResult>;`
- `fixText(text: string, options: TextlintKernelOptions): Promise<TextlintFixResult>;`

## Proposal

This proposal proposes a more primitive API.

- `applyMessages({ text: string, messages: TextlintMessage[] }): Promise<{ text: string })>`

It just exposes `sequentiallyApplyFixes` as public API. 

## Why

Currently, the external user like editor tool can not interop into fixing text.

Editor app like @textlint/editor do the following workflow

1. Lint text → get `TextlintResult`
2. Show `TextlintResult`'s `messages` as errors on the editor
3. Fix specific error messages

textlint does not expose "Fix specific error messages" API yet.
`fixText` API fix all text with all rules.

workaround: just text replace.

```
text.slice(0, message.fix.range[0]) + message.fix.text + text.slice(message.fix.range[1])
```

However, this workaround can not support multiple messages once.
This reason is that I've created `sequentiallyApplyFixes`.

### Another Approach

Split out `sequentiallyApplyFixes`  into a library.

## UseCase

- The user can fix text with a specific mesage
- The user can fix text with specific multiple mesages
   - In other words, The user can fix text with specific rule
- Allow to implement filtering for messages externally
   - @textlint/editor want to implement in-memory ignoring messages

azu:
split out `source-code-fixer.ts` to `@textlint/source-code-fixer`

I have created a new package instead of adding a new API into `@textlint/kernel`.
If make fixer stable, we can move it to kernel public API.

fix #735


packages/@textlint/source-code-fixer/README.md
```diff
@@ -0,0 +1,56 @@
+# @textlint/source-code-fixer
+
+Apply @textlint/kernel lint result to text.
+
+Almost user just user `@textlint/kernel`'s `fixText` API instead of This package.
```
azu, line 5: `fixText` is a stable API.


If you want to get fixed results, please just use kernel API.

Issue #746, opened 2021-03-19
export broken xml with checkstyle format


chick-p:
**What version of textlint are you using?**

v11.8.2

**What file type (Markdown, plain text, etc.) are you using?**

Markdown

**What did you do? Please include the actual source code causing the issue.**

If rule ID has XML control characters, the checkstyle result is broken.
I've used with [textlint-rule-preset-JTF-style](https://github.com/textlint-ja/textlint-rule-preset-JTF-style) preset.
A target file has an NG pattern as `4.3.7.山かっこ<>`.

Target file: 

```markdown
# README

原則として和文では山かっこ < は使用しない。
```

Result:

```shell
$ npx textlint --preset textlint-rule-preset-jtf-style --format checkstyle README.md
<?xml version="1.0" encoding="utf-8"?><checkstyle version="4.3"><file name="/Users/chick-p/textlint-playground/README.md"><error line="3" column="15" severity="error" message="&lt;の対となる&gt;が見つかりません。&lt;&gt; (jtf-style/4.3.7.山かっこ<>)" source="eslint.rules.jtf-style/4.3.7.山かっこ&lt;&gt;" /></file></checkstyle>
```

If use reviewdog, cannot pass the textlint result.

```
reviewdog: parse error: XML syntax error on line 3: unescaped < inside quoted string
error Command failed with exit code 1.
Error: Process completed with exit code 1.
```

**What did you expect to happen?**

Even if a rule ID has XML control characters, I'd like to make textlint escapes the rule ID.

**What actually happened? Please include the actual, raw output from textlint.**

```
$ npx textlint --preset textlint-rule-preset-jtf-style --format checkstyle README.md --debug
  textlint:cli textlint --version: 11.8.2 +0ms
  textlint:cli Running on files, stdin-filename: undefined +2ms
  textlint:module-loader config Config {
  textlint:module-loader   configFile: undefined,
  textlint:module-loader   rulesBaseDirectory: undefined,
  textlint:module-loader   disabledRules: [],
  textlint:module-loader   rules: [],
  textlint:module-loader   disabledFilterRules: [],
  textlint:module-loader   filterRules: [],
  textlint:module-loader   presets: [ 'jtf-style' ],
  textlint:module-loader   plugins: [],
  textlint:module-loader   pluginsConfig: {},
  textlint:module-loader   rulesConfig:
  textlint:module-loader    { 'jtf-style/1.1.1.本文': true,
[...]

PR #747, opened 2021-03-19
fix(linter-formatter): escape  XML control characters in ruleId

chick-p:
Please refer to #746.

This PR is make escape XML control characters in ruleId.


azu:
Nice catch! Thanks!

PR #757, opened 2021-04-25
feat: support sync API

### LINKED ISSUE #756: [feat] Can you please provide `sync` API with `async` together?
All APIs changed to be `async` in #75, but it would be great to still support legacy `sync` APIs.

I'm trying to integrate this tool to ESLint which does not support `async` rule for now.

So my proposal is to add `lintTextSync`, `fixTextSync` APIs, etc.

Some async rules like `textlint-rule-no-dead-link` won't work for `sync` API, that's OK, the user should just disable `async` rules if they want to use `sync` API.

If you agree, I'm glad to raise a PR for it.

JounQin:
close #756

I'm just copying and pasting `sync` APIs for review first, it seems much easier than I imagined and tried on [markuplint](https://github.com/markuplint/markuplint/pull/155), and it inspires me to simplify the `markuplint` PR.

cc @azu


JounQin:
@azu 

Now that I've [implemented](https://github.com/rx-ts/eslint/pull/3) to use markuplint async API but as sync via `child_process` which is inspired by [sync-threads](https://github.com/lambci/sync-threads#appendix) actually, this PR can be closed if you really don't want this feature.

But I still want to say that it would be great to support sync API for performance reason like my integration. (serialization/deserialization between calls and io costing for files)

azu:
Thanks for the work!

These are different implementations

Almost library that has sync and async API uses the same logic in sync/async.

abstruction code
```js
syncAPI(){
  // logic
}
async asyncAPI(){
  return Promise.resolve(syncAPI());
}
```

However, this implementation has two logic for sync and async.
It increases maintenance costs. 
It is hard that we maintenance both sync and async, I think.
https://github.com/textlint/textlint/issues/756#issuecomment-826037987

So, I've closed this PR and issue.

Thanks again for implementing PoC.

PR #783, opened 2021-05-23
feat(ast-tester): improve error message

azu:
- kernel does not use ast-tester. Instead of it, detect the pluginParsedObject
- add tests for #649 when --fix mode
- mitigate the AST check and it improve perf


packages/@textlint/kernel/src/fixer/fixer-processor.ts
```diff
@@ -69,9 +69,9 @@ export default class FixerProcessor {
             return (sourceText: string): Promise<string> => {
                 // create new SourceCode object
                 const preProcessResult = preProcess(sourceText, sourceCode.filePath);
-                const isPluginReturnAnAST = isTxtAST(preProcessResult);
-                const textForAST = isPluginReturnAnAST ? sourceText : preProcessResult.text;
-                const ast = isPluginReturnAnAST ? preProcessResult : preProcessResult.ast;
+                const isParsedObject = isPluginParsedObject(preProcessResult);
```
azu, line 72: mitigate checking of the AST that is returned by plugin.

It aims to improve perf.

PR #793, opened 2021-05-31
Use native map instead of map-like

### LINKED ISSUE #792: use native `Map` instead or `map-like`
https://github.com/textlint/textlint/search?q=map-like

We have used [map-like](https://github.com/azu/map-like) for compatibility ES5 env.
However, we have dropped ES5 env supports in [textlint v12.0.0](https://textlint.github.io/blog/2021/05/24/textlint-12.html).

So, We can just use `Map` instead or `map-like`.

## Task

- [ ] use native `Map` instead or `map-like`
- [ ] Remove `ma-like` dependencies

RhnSharma:
Fixes #792
Hi @azu, how does this look?
Let me know if this needs any changes.
Thanks

### DIFF
--- a/packages/@textlint/feature-flag/package.json
+++ b/packages/@textlint/feature-flag/package.json
@@ -35,9 +35,7 @@
     "test": "mocha \"test/**/*.ts\"",
     "watch": "tsc -b --watch"
   },
-  "dependencies": {
-    "map-like": "^2.0.0"
-  },
+  "dependencies": {},
   "devDependencies": {
     "@types/mocha": "^8.2.2",
     "@types/node": "^14.17.1",
--- a/packages/@textlint/feature-flag/src/index.ts
+++ b/packages/@textlint/feature-flag/src/index.ts
@@ -1,8 +1,7 @@
 // MIT © 2017 azu
 "use strict";
-import { MapLike } from "map-like";
 
-const flagMap = new MapLike<string, boolean>();
+const flagMap = new Map<string, boolean>();
 /**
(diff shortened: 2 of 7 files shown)


azu (commented):
Thanks for PR!
We need to update `yarn.lock`, please run `yarn install` on root dir.

azu:
```
Error: src/engine/rule-map.ts(13,28): error TS2339: Property 'length' does not exist on type 'IterableIterator<string>'.
Error: src/engine/rule-map.ts(40,27): error TS2345: Argument of type 'Function | undefined' is not assignable to parameter of type 'Function'.
```

Some types occur.

> Error: src/engine/rule-map.ts(13,28): error TS2339: Property 'length' does not exist on type 'IterableIterator<string>'.

Probably,  we can resolve it by replacing length with  [Map.prototype.size](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/size) instead of length.

```diff
    hasRuleAtLeastOne() {
-     return this.keys().length > 0
+     return this.size > 0
    }
```      

> Error: src/engine/rule-map.ts(40,27): error TS2345: Argument of type 'Function | undefined' is not assignable to parameter of type 'Function'.

Probably, we can resolve it by removing `| undefined`


```diff
-     defineRule(ruleKey: string, ruleHandler: Function | undefined) {
+    defineRule(ruleKey: string, ruleHandler: Function) {
```

azu:
All tests passed 🎉 

Thanks!

PR #797, opened 2021-06-30
chore: export more symbols to public API

JounQin:
In `@markuplint/textlint`, we're using some symbols but not exported from `textlint` and `@textlint/kernel`'s entry.

https://github.com/markuplint/markuplint/blob/main/packages/%40markuplint/rule-textlint/src/helper.ts#L2

https://github.com/markuplint/markuplint/blob/main/packages/%40markuplint/rule-textlint/src/verify.ts#L2

---

Besides, maybe `export *` can be used.

### DIFF
--- a/packages/textlint/src/index.ts
+++ b/packages/textlint/src/index.ts
@@ -26,3 +26,6 @@ export { TextFixEngine } from "./textfix-engine";
 
 // Core API for linting a **single** text or file.
 export { TextLintCore } from "./textlint-core";
+
+// Config API for loading textlint config
+export { Config } from "./config/config";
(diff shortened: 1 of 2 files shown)


packages/textlint/src/index.ts
```diff
@@ -26,3 +26,6 @@ export { TextFixEngine } from "./textfix-engine";
 
 // Core API for linting a **single** text or file.
 export { TextLintCore } from "./textlint-core";
+
+// Config API for loading textlint config
+export { Config } from "./config/config";
```
azu, line 31: Config is unstable API. We not want to export it.
(It will be changed in the future)

JounQin: Well, it's already a public API by `import { Config } from 'textlint/lib/config/config';`

But after upgrading to v12, there are `textlint/lib/src/config/config` for cjs vs `textlint/module/src/config/config` for esm.

It's just bad for a consumer.

Maybe `exports` field in `package.json` should be used.

azu: Yes. It is internal API.

PR #805, opened 2021-07-26
Add 校正さん to user list

hata6502:
from: https://textlint.github.io/users.html

Would you like to add 校正さん to user list?
校正さん is a note application with proofreading.
Textlint is used on browser.

detail: https://kohsei-san.b-hood.site/lp/

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -34,6 +34,12 @@ const users = [
         image: "/img/soracom.png",
         infoLink: "https://soracom.jp",
         pinned: true
+    },
+    {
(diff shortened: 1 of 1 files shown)


azu:
Thanks

0x6b:
@hata6502 I just find that https://kohsei-san.b-hood.site/ is not found. Is it intended? Should we remove your site from the user list, or wait for coming back?

hata6502:
I relaunched 校正さん with another app domain. 
https://github.com/textlint/textlint/pull/910

PR #773, opened 2021-05-22
Drop to support Node.js 10.x

azu:
- Update TypeScript 3.3
- Update `engines` field
- Update CI - drop node 10 and add node 16

fix #600

PR #814, opened 2021-09-04
chore: remove the husky config and update the related docs

shuuji3:
`husky` was removed from this project in this commit: [Drop to support Node.js 10.x by azu · Pull Request #773 · textlint/textlint](https://github.com/textlint/textlint/pull/773/commits/f25da7e91ff30f473287c9c4d69946f4750bc966#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L43) so I believe we don't need husky related config anymore.

I found a husky-related description in `docs/CONTIRIBUTING.md` and updated it. Also, `npm run prettier` was replaced with `npm run format` by the same commit. I updated that command too.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -141,7 +141,7 @@ We are migrating entire codes to TypeScript.
 
 ##### Linting and Style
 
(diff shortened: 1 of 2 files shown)


azu:
Nice catch!
Thanks!

PR #828, opened 2021-11-19
Convert textlint package's tests and fixtures to TS

pjtsearch:
## Converted Tests/Fixtures

- textlint
- config
- cli
- config-loader
- execute-file-backers
- object-to-kernel-format
- other
- plugin-loader
- plugins
- rule-context
- rule-tips
- textlint-core-fixer
- textlint-module-mapper
- textlint-module-resolver

## Refactorings needed for conversion

- Have ```loadFromDir``` be able to filter multiple extensions and load JS and TS by default so that ```loadFromConfig``` will load TS fixtures
- Broaden TextLintModuleMapper ```createEntities``` and ```createMappedObject``` so that can
apply to ```rules``` and ```rulesConfig``` obejcts
- Refactored all rule fixtures so they use computed property names

Ref #369


azu:
Wow, great work!
 I'll look it tonight.

packages/textlint/src/engine/rule-loader.ts
```diff
@@ -19,10 +19,16 @@ export function loadFromDir(rulesDir: string, extname: string = ".js"): { [index
...
+            if (!extnames.includes(path.extname(file))) {
+                return;
+            }
+        } else {
+            if (path.extname(file) !== extnames) {
+                return;
+            }
         }
-        const withoutExt = path.basename(file, extname);
+        const withoutExt = path.basename(file, path.extname(file));
```
azu, line 31: We can declare `fileExtname` at first for reducing duplicate.

```ts
const fileExtname = path.extname(file);
if (Array.isArray(extnames)) {
    if (!extnames.includes(fileExtname)) {
        return;
    }
} else {
    if (fileExtname !== extnames) {
        return;
    }
}
const withoutExt = path.basename(file, fileExtname);
```

packages/textlint/test/config-loader/fixtures/@textlint/textlint-config-example/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "textlint-config-example",
   "version": "1.0.0",
   "description": "",
-  "main": "index.js",
+  "main": "index.ts",
```
azu, line 5: It is a bit hacky, but it is reasonable.

azu (approved):
LGTM!

Thanks for nice work!

PR #831, opened 2021-11-29
Replace isTextlintAST with isTxtAST

alpaca-tc:
Fix typo documentation. 

🙅 `isTextlintAST` is wrong
🙆‍♂️ `isTxtAST` is correct.

https://github.com/textlint/textlint/blob/abb75bb331a5cd563934cce362cea62d6de740ac/packages/%40textlint/ast-tester/src/index.ts#L14


azu:
You're correct! Thanks!

PR #832, opened 2021-12-01
Plugin can return Promise when calling preProcess

alpaca-tc:
## What's changed

- Promise type added to type definition of `preProcess`.
- The method from which preProcess is called is now an `async` function. Fortunately, it was originally a function that returned a Promise, so the change is minimal.

## Why make this change?

In [textlint-plugin-ruby](https://github.com/alpaca-tc/textlint-plugin-ruby), which I'm developing now, it communicates with the backend ruby process to parse the Ruby source code.
However, since node IO is mainly asynchronous, it is difficult to implement unless preProcess supports the Promise type 😭 

---

By the way, thanks for the replies on [twitter](https://twitter.com/azu_re/status/1465771946675945472?s=20)! 💛

### DIFF
--- a/packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
+++ b/packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
@@ -40,7 +40,10 @@ export declare class TextlintPluginProcessor {
          * @param text
          * @param filePath
          */
-        preProcess(text: string, filePath?: string): TxtNode | { text: string; ast: TxtNode };
+        preProcess(
+            text: string,
+            filePath?: string
+        ): TxtNode | { text: string; ast: TxtNode } | Promise<TxtNode> | Promise<{ text: string; ast: TxtNode }>;
         postProcess(messages: Array<any>, filePath?: string): { messages: Array<any>; filePath: string };
     };
 }
--- a/packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
+++ b/packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
@@ -10,6 +10,7 @@ export declare type TextlintPluginOptions = {
 };
 
 export type TextlintPluginPreProcessResult = TxtNode | { text: string; ast: TxtNode };
+export type TextlintPluginPostProcessResult = { messages: Array<any>; filePath: string };
 
 export interface TextlintPluginProcessorConstructor extends Function {
     new (options?: TextlintPluginOptions): TextlintPluginProcessor;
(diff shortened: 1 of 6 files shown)


packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
```diff
@@ -40,7 +40,10 @@ export declare class TextlintPluginProcessor {
          * @param text
          * @param filePath
          */
-        preProcess(text: string, filePath?: string): TxtNode | { text: string; ast: TxtNode };
+        preProcess(
+            text: string,
+            filePath?: string
+        ): TxtNode | { text: string; ast: TxtNode } | Promise<TxtNode> | Promise<{ text: string; ast: TxtNode }>;
```
azu, line 46: Maybe,

- Add `PreProcessResullt` type
- Return `PreProcessResullt | Promise<PreProcessResullt>`

It is more reasonable.

```ts
type PreProcessResult = TxtNode | { text: string; ast: TxtNode };
...
        preProcess(
            text: string,
            filePath?: string
        ): PreProcessResullt | Promise<PreProcessResullt>;
```

azu (commented):
Look like good about `preProcess`.

Can you add same change to `postProcess`?
Also, if you can, please add a test case into https://github.com/textlint/textlint/blob/master/packages/%40textlint/kernel/test/kernel-plugin-test.ts


- Create `AsyncExamplePlugin` like follows
- Add a test case using `AsyncExamplePlugin`

```ts
import { TextlintMessage } from "@textlint/kernel";
import type { TextlintPluginCreator, TextlintPluginProcessor } from "@textlint/types";
import { parse } from "@textlint/markdown-to-ast";

export interface ExampleProcessorOptions {
    testOption: string;
}

export interface CreateAsyncPluginOptions {
    extensions?: string[];
}

/**
 * Create Async Plugin stub
 * It spy the assigned argument.
 * It is compatible with markdown plugin by default.
 */
export const createAsyncPluginStub = (options?: CreateAsyncPluginOptions) => {
    let assignedOptions: undefined | {};
    let processorArgs: {
        extension: string;
    };
    let preProcessArgs: {
        text: string;
        filePath: string;
    };
    let postProcessArgs: {
        messages: TextlintMessage[];
        filePath?: string;
    };
    return {
        getOptions() {
            return assignedOptions;
        },
        getProcessorArgs() {
            return processorArgs;
        },
        getPreProcessArgs() {
            return preProcessArgs;
        },
        getPostProcessArgs() {
            return postProcessArgs;
        },
        /**
         * Return plugin module
         */
        get plugin(): TextlintPluginCreator {
            return {
                Processor: class MockExampleProcessor implements TextlintPluginProcessor {
                    availableExtensions() {
                        return (options && options.extensions) || [".md"];
                    }

                    constructor(public options?: {}) {
                        assignedOptions = options;
                    }

                    processor(extension: string) {
                        processorArgs = { extension };
                        return {
                            async preProcess(text: string, filePath: string) {
                                preProcessArgs = { text, filePath };
                                return parse(text);
                            },
                            async postProcess(messages: TextlintMessage[], filePath: string) {
                                postProcessArgs = { messages, filePath };
                                return {
                                    messages,
                                    filePath: filePath || "unknown"
                                };
                            }
                        };
                    }
                }
            };
        }
    };
};
```

alpaca-tc:
@azu Thank you for reviewing 💛

azu (approved):
LGTM

azu:
Thanks!
We have released it as https://github.com/textlint/textlint/releases/tag/v12.1.0

Issue #840, opened 2022-02-21
Update ansi-regex


massongit:
`ansi-regex` 2.1.1 ~ 5.0.0 is vulnerable to Inefficient Regular Expression Complexity: CVE-2021-3807
Therefore, you need to update it to 5.0.1.

[Packages that need to update]

* [x] `chalk` (Update to v4) #842 
* [ ] `log-symbols` #845 
* [x] `mocha`
* [x] `pkg-to-readme` (in `textlint-scripts` )
* [x] `string-width` #844 

[Packages that do not update `ansi-regex` to 5.0.1 after updating to the latest version]
* `docusaurus` (in `textlint-website` )
* `gulp` (in `gulp-textlint` )
* `lerna`

azu:
Thanks for notice, a PR is welcome!

Probably, we need to update chalk@5, but chalk does not provice CommonJS.
[chalk@5](https://github.com/chalk/chalk/releases/tag/v5.0.0) provice only ESM.

So, we need to update to chalk@4.

```
yarn list v1.22.17
├─ @textlint/fixer-formatter@12.1.0
│  ├─ ansi-regex@5.0.1
│  └─ chalk@1.1.3
│     └─ ansi-regex@2.1.1
├─ @textlint/linter-formatter@12.1.0
│  ├─ ansi-regex@5.0.1
│  └─ chalk@1.1.3
│     └─ ansi-regex@2.1.1
├─ ansi-regex@2.1.1
├─ cli-truncate@2.1.0
│  └─ ansi-regex@5.0.0
├─ cliui@7.0.4
│  └─ ansi-regex@5.0.0
├─ eslint@8.6.0
│  └─ ansi-regex@5.0.1
├─ inquirer@7.3.3
│  └─ ansi-regex@5.0.0
├─ log-update@4.0.0
│  └─ ansi-regex@5.0.0
├─ react-dev-utils@11.0.4
│  └─ ansi-regex@5.0.0
├─ table@6.7.5
│  └─ ansi-regex@5.0.1
├─ wide-align@1.1.3
│  └─ ansi-regex@3.0.0
├─ wrap-ansi@7.0.0
│  └─ ansi-regex@5.0.0
└─ yargs@16.2.0
   └─ ansi-regex@5.0.0
✨  Done in 0.42s.
```

azu:
[...]

PR #841, opened 2022-02-22
Update packages that depend on ansi-regex

massongit:
https://github.com/textlint/textlint/issues/840

`ansi-regex` 2.1.1 ~ 5.0.0 is vulnerable to Inefficient Regular Expression Complexity: CVE-2021-3807
Therefore, I update packages that depend on `ansi-regex` that can be updated without source code modifications.


massongit:
The following packages are excluded from this PR as they fail testing when updated.
* `chalk`
* `string-width`
* `strip-ansi`

azu:
Thanks

Issue #843, opened 2022-02-23
Update string-width@4


azu:
> After check again, it seems that it is not necessary to update `string-width` to update `ansi-regex`.
Therefore, it was excluded from the update target.

_Originally posted by @massongit in https://github.com/textlint/textlint/pull/842#discussion_r812521858_

## TODO

- [ ] Update to string-widtth@4
  - string-width@5 provide only ESM

azu:
This updates does not includes API changes.
https://github.com/sindresorhus/string-width/compare/v1.0.2...v2.0.0
https://github.com/sindresorhus/string-width/releases/tag/v3.0.0
https://github.com/sindresorhus/string-width/releases/tag/v4.0.0

So, we can just update it.

PR #844, opened 2022-02-23
Update string-width

massongit:
fix https://github.com/textlint/textlint/issues/843

I update `string-width` to 4.2.3.


azu:
Thanks

PR #848, opened 2022-02-23
CI: enable cache

azu:
- Enable Cache
  - https://github.com/actions/setup-node
- remove duplicated build task


.github/workflows/ci.yml
```diff
@@ -22,6 +22,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: ${{ matrix.node }}
+          cache: 'yarn'
```
azu, line 25: https://github.com/actions/cache#cache-limits
Cache limit 10GB.
current repository size ~1GB

PR #861, opened 2022-04-30
Update link to collection of textlint rules

parkerbxyz:
This fixes the link to the Wiki page that contains a collection of textlint rules.


azu:
https://github.com/textlint/textlint/wiki/Collection-of-textlint-rule
I've reverted the wiki URL.

PR #895, opened 2022-09-04
docs(readme): delete --global in readme Closes #885

Uda-Titor:
resolve https://github.com/textlint/textlint/issues/885

Issue #855, opened 2022-04-01
Textlint nags about .html in links


villesau:
**What version of textlint are you using?**

The one in superlint 4.9.1:

    "textlint": "^12.1.1",
    "textlint-filter-rule-allowlist": "^4.0.0",
    "textlint-filter-rule-comments": "^1.2.2",
    "textlint-rule-terminology": "^2.1.5",

**What file type (Markdown, plain text, etc.) are you using?**

Markdown (github)

**What did you do? Please include the actual source code causing the issue.**

I have following markdown:

```
## some links

https://forsetisecurity.org/docs/latest/setup/index.html
```

**What did you expect to happen?**

Textlint should not nag

**What actually happened? Please include the actual, raw output from textlint.**

```
2022-04-01 13:22:37 [INFO]   
2022-04-01 13:22:37 [INFO]   ----------------------------------------------
2022-04-01 13:22:37 [INFO]   ----------------------------------------------
2022-04-01 13:22:37 [INFO]   Linting [NATURAL_LANGUAGE] files...
2022-04-01 13:22:37 [INFO]   ----------------------------------------------
2022-04-01 13:22:37 [INFO]   ----------------------------------------------
2022-04-01 13:22:37 [INFO]   ---------------------------
2022-04-01 13:22:37 [INFO]   File:[/github/workspace/README.md]
2022-04-01 13:22:38 [ERROR]   Found errors in [textlint] linter!
2022-04-01 13:22:38 [ERROR]   Error code: 1. Command output:
------

/github/workspace/README.md
  60:53  ✓ error  Incorrect usage of the term: “html”, use “HTML” instead  terminology

✖ 1 problem (1 error, 0 warnings)
✓ 1 fixable problem.
Try to run: $ textlint --fix [file]
```

So textlint thinks links as any text. However, links should not be checked by textlint.

azu:
This is third party issue.
Please create an issue to https://github.com/sapegin/textlint-rule-terminology

villesau:
Looks like there already was an issue about this: https://github.com/sapegin/textlint-rule-terminology/issues/38

PR #899, opened 2022-09-11
docs: replace `./node_modules/.bin/` with `npx`

chalkygames123:
At the risk of nitpicking, I prefer `npx` introduced in npm 5.2 (released in July 2017) over `./node_modules/.bin/` here.


azu:
Related #895 #855

PR #904, opened 2022-09-20
fix(markdown-to-ast): add `mdast-util-gfm-autolink-literal` to explicit deps

### LINKED ISSUE #903: Missing dependency from @textlint/markdown-to-ast to mdast-util-gfm-autolink-literal
This require statement at HEAD
https://github.com/textlint/textlint/blob/a69c0a79a371a64117b21287b41603dd193b9624/packages/%40textlint/markdown-to-ast/src/parse-markdown.ts#L3
depends on a library that's not a declared dependency:
https://github.com/textlint/textlint/blob/a69c0a79a371a64117b21287b41603dd193b9624/packages/%40textlint/markdown-to-ast/package.json#L37-L46

It only happens to work for some users because that dependency is hoisted by the package manager to a place that node resolves it.
This can be reproduced by adding `hoist=false` to `.npmrc` and running under `pnpm`.

azu:
pnpm or yarn 2+ require explicit `dependencies`.

fix #903

### DIFF
--- a/packages/@textlint/markdown-to-ast/package.json
+++ b/packages/@textlint/markdown-to-ast/package.json
@@ -41,6 +41,7 @@
     "remark-frontmatter": "^3.0.0",
     "remark-gfm": "^1.0.0",
     "remark-parse": "^9.0.0",
+    "mdast-util-gfm-autolink-literal": "^0.1.0",
     "traverse": "^0.6.6",
     "unified": "^9.2.2"
   },


packages/@textlint/markdown-to-ast/package.json
```diff
@@ -41,6 +41,7 @@
     "remark-frontmatter": "^3.0.0",
     "remark-gfm": "^1.0.0",
     "remark-parse": "^9.0.0",
+    "mdast-util-gfm-autolink-literal": "^0.1.0",
```
azu, line 44: curretly, use 0.1.0.
This is unclear version because remark-gfm depened on this.
https://github.com/remarkjs/remark-gfm/blob/6161ecc1cc9e79f247355d5ca165f555e934b272/package.json#L36
https://github.com/syntax-tree/mdast-util-gfm/blob/fdf2d47894d3b04db219eaf2e635d9745e555e32/package.json#L38-L44

azu: https://github.com/textlint/textlint/blob/a69c0a79a371a64117b21287b41603dd193b9624/yarn.lock#L10814-L10823

Issue #933, opened 2022-11-10
@textlint/linter-formatter: missing dependencies: lodash


takenspc:
**What version of textlint are you using?**

v12.2.2

**What file type (Markdown, plain text, etc.) are you using?**

Markdown

**What did you do? Please include the actual source code causing the issue.**

I want to use junit formatter:

```sh
npx textlint -f junit README.md
```

**What did you expect to happen?**

JUnit XML will be displayed.

**What actually happened? Please include the actual, raw output from textlint.**

Following error messages are displayed:

```
Error
Could not find formatter junit
Error: Cannot find module 'lodash'
Require stack:
- (snip)/node_modules/@textlint/linter-formatter/lib/src/formatters/junit.js
- (snip)/node_modules/@textlint/linter-formatter/lib/src/index.js
- (snip)/node_modules/textlint/lib/src/options.js
- (snip)/node_modules/textlint/lib/src/cli.js
- (snip)/node_modules/textlint/bin/textlint.js

Stack trace
Error: Could not find formatter junit
Error: Cannot find module 'lodash'
Require stack:
- (snip)/node_modules/@textlint/linter-formatter/lib/src/formatters/junit.js
- (snip)/node_modules/@textlint/linter-formatter/lib/src/index.js
- (snip)/node_modules/textlint/lib/src/options.js
- (snip)/node_modules/textlint/lib/src/cli.js
- (snip)/node_modules/textlint/bin/textlint.js
    at createFormatter ((snip)/node_modules/@textlint/linter-formatter/lib/src/index.js:40:15)
    at TextLintEngine.onFormat ((snip)/node_modules/textlint/lib/src/textlint-engine.js:48:59)
    at TextLintEngine.formatResults ((snip)/node_modules/textlint/lib/src/engine/textlint-engine-core.js:239:32)
    at (snip)/node_modules/textlint/lib/src/cli.js:221:39
```

azu:
Thanks for report.
https://github.com/textlint/textlint/blob/master/packages/%40textlint/linter-formatter/package.json
We should add `lodash` to dependencies

PR #934, opened 2022-11-10
chore(linter-formatter): added lodash to dependencies

takenspc:
... and removed xml-escape which is not used.

fixes https://github.com/textlint/textlint/issues/933.


azu:
Thanks!

Fixed in [v12.2.3](https://github.com/textlint/textlint/releases/tag/v12.2.3)

PR #944, opened 2022-11-22
chore(ast-node-types): add `children` to `TxtNode`

unvalley:
Thanks for great tool, this PR doesn't relate any issues.

Is there any reason `TxtNode` does not have `children` property?
https://github.com/textlint/textlint/blob/312b2d579ed3bebd786f9674605395458403e55e/packages/%40textlint/ast-node-types/src/index.ts#L82
I'd like to use `text-to-ast` to parse text to AST, and access the node `children` but currently, I guess accessing the node `children` isn't type safe. (`TxtNode["children"]` returns `any`).

### DIFF
--- a/packages/@textlint/ast-node-types/src/index.ts
+++ b/packages/@textlint/ast-node-types/src/index.ts
@@ -84,10 +84,11 @@ export interface TxtNode {
     raw: string;
     range: TextNodeRange;
     loc: TxtNodeLineLocation;
-    // parent is runtime information
+    // parent and children is runtime information
     // Not need in AST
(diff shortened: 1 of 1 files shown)


azu:
https://github.com/textlint/textlint/blob/1baa72ad2809efe90044175f9bd0b35e18251c7b/packages/@textlint/ast-node-types/src/TypeofTxtNode.ts#L63-L74

My memory is a bit fuzzy, but `TxtNode` is used as a Node for self-closed tags.
It seems a little strange that TxtNode is being reused.
Perhaps it would be better to add a new TxtSelfClosedNode and redefine TxtNode.

Issue #996, opened 2023-01-13
Remove `events` modules from @textlint/kernel


azu:
- #985 

We want to remove Node.js built-in module from `@textlint/kernel/src/*`.

`assert` has been removed in #985 by @Sean0628
`events` is still used.

```
❯ npx find-node-modules-import "src/**/*.ts" --builtinModules
src/task/promise-event-emitter.ts:5:30	events
src/task/textlint-core-task.ts:9:30	events
```

- https://github.com/textlint/textlint/blob/cd91d67eb089457ebebf7560bd9f2cb6015a2867/packages/@textlint/kernel/src/task/promise-event-emitter.ts#L4
- https://github.com/textlint/textlint/blob/cd91d67eb089457ebebf7560bd9f2cb6015a2867/packages/%40textlint/kernel/src/task/textlint-core-task.ts#L9

Both use `on`/`emit`/`listenerCount`.

Probably, We can rewrite `events` as own module like this

```js
// Example EventEmitter https://jsprimer.net/use-case/todoapp/event-model/#event-emitter
export class EventEmitter {
    // 登録する [イベント名, Set(リスナー関数)] を管理するMap
    #listeners = new Map();
    /**
     * 指定したイベントが実行されたときに呼び出されるリスナー関数を登録する
     * @param {string} type イベント名
     * @param {Function} listener イベントリスナー
     */
    addEventListener(type, listener) {
        // 指定したイベントに対応するSetを作成しリスナー関数を登録する
        if (!this.#listeners.has(type)) {
            this.#listeners.set(type, new Set());
        }
        const listenerSet = this.#listeners.get(type);
        listenerSet.add(listener);
    }

    listenerCount(eventName){
        return this.#listeners.get(type)?.size ?? 0;
    } 

    /**
     * 指定したイベントをディスパッチする
     * @param {string} type イベント名
     */
    emit(type) {
        // 指定したイベントに対応するSetを取り出し、すべてのリスナー関数を呼び出す
        const listenerSet = this.#listeners.get(type);
        if (!listenerSet) {
            return;
        }
        listenerSet.forEach(listener => {
            listener.call(this);
        });
    }

    /**
     * 指定したイベントのイベントリスナーを解除する
     * @param {string} type イベント名
     * @param {Function} listener イベントリスナー
     */
    removeEventListener(type, listener) {
        const listenerSet = this.#listeners.get(type);
        if (!listenerSet) {
            return;
        }
        listenerSet.forEach(ownListener => {
[...]

Issue #730, opened 2020-12-21
TIMING=1 does not work correctly


azu:
Probably, TIMING=1 does not work on the async rule.

We should fix it.
https://textlint.github.io/docs/rule.html#rule-performance

PR #1003, opened 2023-01-22
refactor(kernel): remove node:events 

azu:
fix https://github.com/textlint/textlint/issues/996
fix https://github.com/textlint/textlint/issues/730

### DIFF
--- a/packages/@textlint/kernel/src/util/timing.ts
+++ b/packages/@textlint/kernel/src/util/timing.ts
@@ -58,7 +58,7 @@ function display(data: any) {
         .slice(0, 10);
 
     rows.forEach(function (row: any) {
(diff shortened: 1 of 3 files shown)


packages/@textlint/kernel/src/util/timing.ts
```diff
@@ -114,9 +114,9 @@ export default (function () {
             data[key] = 0;
         }
 
-        return function () {
+        return async function (...args: any[]) {
             let t = process.hrtime();
-            fn.apply(null, Array.prototype.slice.call(arguments));
+            await fn(...args);
```
azu, line 119: This may fix https://github.com/textlint/textlint/issues/730

PR #1033, opened 2023-01-29
fix(ast-node-types): remove markdown extension && fix TxtNodeType

azu:
- Correct `type` value in some Nodes
- Remove Markdown Extension from ast-node-types
- Remove `string` from `TxtNodeType`
  - It make `TxtNodeType` as `string`
- `_shouldNotUsed` use `never` type

### DIFF
--- a/packages/@textlint/ast-node-types/src/NodeType.ts
+++ b/packages/@textlint/ast-node-types/src/NodeType.ts
@@ -12,7 +12,7 @@ import type { ASTNodeTypes } from "./ASTNodeTypes";
  * Key of ASTNodeTypes or any string
  * For example, TxtNodeType is "Document".
  */
-export type TxtNodeType = keyof typeof ASTNodeTypes | string;
+export type TxtNodeType = keyof typeof ASTNodeTypes;
 
 /**
  * Any TxtNode types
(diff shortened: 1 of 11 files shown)


packages/@textlint/types/src/Rule/TextlintRuleContext.ts
```diff
@@ -23,5 +23,5 @@ export type TextlintRuleContextReportFunction = (args: TextlintRuleContextReport
 
 export interface TextlintRuleContext extends TextlintBaseRuleContext {
     fixer: TextlintRuleContextFixCommandGenerator;
-    report: (node: TxtNode, ruleError: TextlintRuleReportedObject | TextlintRuleError, _shouldNotUsed?: any) => void;
+    report: (node: TxtNode, ruleError: TextlintRuleReportedObject | TextlintRuleError, _shouldNotUsed?: never) => void;
```
azu, line 26: It should be `never` type

packages/textlint/test/cli/fixtures/fixer-rules/fixer-rule-add.ts
```diff
@@ -6,6 +6,9 @@ import { TextlintRuleContext, TextlintRuleReportHandler } from "@textlint/types"
 const reporter = (context: TextlintRuleContext): TextlintRuleReportHandler => {
     const { Syntax, fixer, report, getSource } = context;
     return {
+        ["test"](_node) {
```
azu, line 9: custom node type test

packages/@textlint/ast-node-types/src/NodeType.ts
```diff
@@ -12,7 +12,7 @@ import type { ASTNodeTypes } from "./ASTNodeTypes";
  * Key of ASTNodeTypes or any string
  * For example, TxtNodeType is "Document".
  */
-export type TxtNodeType = keyof typeof ASTNodeTypes | string;
+export type TxtNodeType = keyof typeof ASTNodeTypes;
```
azu, line 15: Its big change.
More type safe

PR #1035, opened 2023-01-29
fix(ast-node-type): fix Node → TxtNode

azu:
`Node` is DOM Node!
It is wrong type.

- `Node` to `TxtNode` 
- Prevent same mistake by `lib`

### DIFF
--- a/packages/@textlint/ast-node-types/tsconfig.json
+++ b/packages/@textlint/ast-node-types/tsconfig.json
@@ -1,7 +1,11 @@
 {
   "extends": "../../../tsconfig.base.json",
   "compilerOptions": {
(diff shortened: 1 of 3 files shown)


packages/@textlint/ast-node-types/tsconfig.json
```diff
@@ -1,7 +1,11 @@
 {
   "extends": "../../../tsconfig.base.json",
   "compilerOptions": {
-    "outDir": "lib"
+    "outDir": "lib",
+    "lib": [
+      "esnext"
+      // "dom"
+    ]
```
azu, line 8: It prevent to reference `Node`

PR #1043, opened 2023-01-31
 test(textlint): remove it.only and use ts-node/esm 

### LINKED ISSUE #1041: Not all tests in textlint have been run

4251c9e

I followed the [Development Workflow
](https://github.com/textlint/textlint/blob/master/docs/CONTRIBUTING.md#development-workflow), added a new test in config-initializer-test.ts and `npm run test` in `packages/textlint`.

 The test I added never failed.

I found mocha's `.only()` is used [here](https://github.com/textlint/textlint/blob/master/packages/textlint/test/textlint-core/textlint-core-test.ts#L58). According to [mocha's doc](https://mochajs.org/#exclusive-tests
), It forces only its specified tests to be executed. 
After removing the `.only`, some tests failed.

azu:
fix #1041


packages/textlint/.mocharc.json
```diff
@@ -1,6 +1,7 @@
 {
-  "timeout": "5000",
-  "require": [
-    "ts-node-test-register"
+  "timeout": 5000,
+  "loader": "ts-node/esm",
```
azu, line 3: Dynamic Import + .ts  require `ts-node/esm`.

PR #1088, opened 2023-02-26
docs: fix parameters of loadTextlintrc

massongit:
```
$ cat test.js
const {loadTextlintrc} = require("textlint");

(async function () {
    const descriptor = await loadTextlintrc();
})()

$ node test.js
/path/to/node_modules/textlint/lib/src/loader/TextlintrcLoader.js:35
const loadTextlintrc = async ({ configFilePath, node_modulesDir }) => {
                                ^

TypeError: Cannot destructure property 'configFilePath' of 'undefined' as it is undefined.
    at loadTextlintrc (/path/to/node_modules/textlint/lib/src/loader/TextlintrcLoader.js:35:33)
    at /path/to/test.js:4:30
    at Object.<anonymous> (/path/to/test.js:5:3)
    at Module._compile (node:internal/modules/cjs/loader:1275:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1329:10)
    at Module.load (node:internal/modules/cjs/loader:1133:32)
    at Module._load (node:internal/modules/cjs/loader:972:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:83:12)
    at node:internal/main/run_main_module:23:47

Node.js v19.7.0
```

I fix parameters of `loadTextlintrc` in a doc because that function needs parameters.

PR #1089, opened 2023-02-26
fix: allow missing parameters for loadTextlintrc

massongit:
https://github.com/textlint/textlint/pull/1088#issuecomment-1445265820

```
$ cat test.js
const {loadTextlintrc} = require("textlint");

(async function () {
    const descriptor = await loadTextlintrc();
})()

$ node test.js
/path/to/node_modules/textlint/lib/src/loader/TextlintrcLoader.js:35
const loadTextlintrc = async ({ configFilePath, node_modulesDir }) => {
                                ^

TypeError: Cannot destructure property 'configFilePath' of 'undefined' as it is undefined.
    at loadTextlintrc (/path/to/node_modules/textlint/lib/src/loader/TextlintrcLoader.js:35:33)
    at /path/to/test.js:4:30
    at Object.<anonymous> (/path/to/test.js:5:3)
    at Module._compile (node:internal/modules/cjs/loader:1275:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1329:10)
    at Module.load (node:internal/modules/cjs/loader:1133:32)
    at Module._load (node:internal/modules/cjs/loader:972:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:83:12)
    at node:internal/main/run_main_module:23:47

Node.js v19.7.0
```

In order to eliminate the above error, I allow missing parameters for `loadTextlintrc`.

### DIFF
--- a/packages/textlint/src/loader/TextlintrcLoader.ts
+++ b/packages/textlint/src/loader/TextlintrcLoader.ts
@@ -37,13 +37,10 @@ export const loadBuiltinPlugins = async (): Promise<TextlintKernelDescriptor> =>
         plugins: builtInPlugins
     });
 };
-export const loadTextlintrc = async ({
-    configFilePath,
-    node_modulesDir
-}: LoadTextlintrcOptions): Promise<TextlintKernelDescriptor> => {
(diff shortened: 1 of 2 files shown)


azu:
Thanks!

PR #1090, opened 2023-02-26
fix(@textlint/linter-formatter, @textlint/fixer-formatter): allow missing parameters for loadFormatter

massongit:
In CLI, default formatter is `stylish`.
Therefore, I also make `stylish` the default formatter in modules.


azu:
Thanks for PR.

I think that the user still should specify formatter option.
Because module user is not CLI user.

`stylish` formatter is suitable for CLI user, but may not be for module user.
The needs of module users are different. Therefore, there is no default formatter.

`textlint` package exports `cli` object that is CLI.
`loadLinterFormatter` is not only for CLI.

Issue #1091, opened 2023-02-26
Fixed result is not shown when both `--dry-run` and `--stdin` are given


kg8m:
Hi. Thanks a lot for developing this great tool. textlint really helps my text-editing life.

* * *

**What version of textlint are you using?**
```sh
$ npx --no-install textlint -v
v13.3.0
```

**What file type (Markdown, plain text, etc.) are you using?**
Markdown (and so on)

**What did you do? Please include the actual source code causing the issue.**
Run the cli textlint as follows:

```sh
$ echo 'jquery' | textlint --fix --dry-run --stdin --stdin-filename test.md --format fixed-result --rule 'textlint-rule-terminology'
```

**What did you expect to happen?**
We see `jQuery\n`.

**What actually happened? Please include the actual, raw output from textlint.**

We see no output.

For your information, the output is shown if removing `--dry-run` option.

I think fixing works properly but its output isn't shown. Here is output with `--debug` option:

```sh
$ echo 'jquery' | textlint --fix --dry-run --stdin --stdin-filename test.md --format fixed-result --rule 'textlint-rule-terminology' --debug
  textlint:cli cliOptions: {"fix":true,"dryRun":true,"stdin":true,"stdinFilename":"test.md","format":"fixed-result","rule":["textlint-rule-terminology"],"debug":true,"ignorePath":"${PWD}/.textlintignore","init":false,"color":true,"quiet":false,"textlintrc":true,"cache":false,"cacheLocation":"${PWD}/.textlintcache","_":[]} +0ms
  textlint:cli textlint --version: 13.3.0 +0ms
  textlint:cli Running on text, stdin-filename: test.md +0ms
  textlint:cli cliDescriptor: {"rule":[{"id":"terminology","options":{}}],"filterRule":[],"plugin":[]} +9ms
[...]

PR #1092, opened 2023-02-26
fix(textlint): fix to print results when the `--dry-run` option is specified.

Sean0628:
This PR attempts to fix to print results when the `--dry-run` option is specified.

fixes: #1091

## TEST
```sh
>>> echo 'jquery' | npx textlint --fix --dry-run --stdin --stdin-filename test.md --format fixed-result --rule 'textlint-rule-terminology'
jQuery
```


azu:
Thanks!

PR #1094, opened 2023-02-26
docs: Add semicolon in code

massongit:
I add a semicolon in the README's code since there is code without a semicolon.


azu:
Thanks

I thought that it was checked by ESLint, but it was not supported by `ts`…
https://github.com/textlint-rule/textlint-rule-eslint/blob/master/src/textlint-rule-eslint.ts#L12

PR #1195, opened 2023-07-07
loadFixerFormatter is not use in example.

KokiSakano:
When I tried this sample code, I did not need `loadFixerFormatter `.So I deleted it.

Also, according to the type definition, `fixText` must require a file path as the second argument, which is correct?

### DIFF
--- a/docs/use-as-modules.md
+++ b/docs/use-as-modules.md
@@ -57,7 +57,7 @@ console.log(output);
 Fix text and get the fixed text.
 
 ```ts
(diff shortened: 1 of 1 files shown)


azu:
`fixText(text, filePath)` is correct.

```
const result = await linter.fixText("TODO: fix me", "DUMMY.md");
```

KokiSakano:
@azu 
Thank you.I fixed it.

azu:
The CI is a bit odd, but the fix is fine.

Thanks!

Issue #1200, opened 2023-07-08
v14 roadmap


azu:
Next major updates lists:

## Summary

I don't think most users will be affected.
If you are using an old Node.js, please update to [Node.js 18+ LTS](https://nodejs.org/en/download/).

- textlint v14 requires Node.js v18.14.0+
- deprecate `TextlintCore`, `TextLintEngine`, and `TextFixEngine`
  - If you use `TextlintCore`, migrate to `@textlint/legacy-textlint-core`. It is based on `@textlint/kernel`, but compatible with `TextlintCore` interface
- textlint-scripts requires Node v18.19.0+ (for dev)
- textlint-scripts output ES2022-comaptible code
- Remove unused CLI and flags

## BREAKING CHANGE

- [x] #1310
  - There are replaced by `createLinter` and `loadTextlintrc`, and `loadLinerFormatter`.
  - [textlint v13.0.0 · textlint](https://textlint.github.io/blog/2023/01/27/textlint-13.html)
- [x] #1317
- [x] Remove a CLI from `@textlint/linter-formatter` #1222 
- [x] Remove experimental `--parallel` flag #633 
- [x] #1336
- [x] #1342
  - `engines` updates
  - related #298   
- [x] #1344
  - Should use native async/await etc...
  - ~~Babel 8?~~

## Features

- [x] add `@textlint/legacy-textlint-core`

azu:
We may remove `--parallel` option. It is not used.

azu:
> Remove deprecated textlint, TextFixEngine, TextLintEngine, and TextLintCore from textllint package

Perhaps the most significant change in this update will be the removal of these.
The ecosystem needs to be investigated to see if this can be safely removed.

azu:
You can try canary version:

```
npm install textlint@next
```

azu:
[Release v14.0.0 · textlint/textlint](https://github.com/textlint/textlint/releases/tag/v14.0.0)

PR #1201, opened 2023-07-08
refactor(textlint): move deprecated code to DEPRECATED/ dir

azu:
- Refs https://github.com/textlint/textlint/issues/1200

Next Major updates will remove deprecated codes.
After this PR, we can just remove `DEPRECATED/` dir.


packages/textlint/src/config/config-initializer.ts
```diff
@@ -66,7 +65,7 @@ export interface CreateConfigFileOption {
 export const createConfigFile = (options: CreateConfigFileOption) => {
     const dir = options.dir;
     return getTextlintDependencyNames(dir).then((pkgNames) => {
-        const rcFile = `.${Config.CONFIG_FILE_NAME}rc.json`;
+        const rcFile = `.textlintrc.json`;
```
azu, line 68: Drop to use deprecated `Config`.

packages/textlint/src/shared/type/MessageType.ts
azu, line 1: Not used

packages/textlint/src/util/fs-promise.ts
azu, line 1: use `fs.promises` instead of it.

packages/textlint/src/util/traverse.ts
azu, line 1: Not used

PR #1295, opened 2023-11-25
fix(ast-node-types): remove `[index:string]: any;` from `TxtNode`

### LINKED ISSUE #1294: Remove `[index:string]: any;` from `TxtNode`.
https://github.com/textlint/textlint/blob/ed732f18a36b3ef7d06dde12aa2e3c39faecabad/packages/@textlint/ast-node-types/src/NodeType.ts#L50-L59

`[index:string]: any;` break `Omit<>` return type.

https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgJLIN4Chm+WATwAcIAuZAIiIoG4c8EALYAGwBMoIRyBnMKUAHMA2gF06eZMNBsIAD1J8BIQaPJwQBOgF8sWQiWQB5ZAF5jAW2BgAPKgA0lAxAoA+ZADJM9XM-IUAe1osXSwEAJA+ZADyI2EKJlYOLgpRM0xtOgB6LMkAPQB+PSA

```ts
interface I {
    type: "p";
    children: string[];
    [index:string]: any;
}

type O = Omit<I, "type"> & {
    type: "o";
}

const o: O["children"] = {};
//    ^? any
```

It prevent that extends TxtNode types.

https://github.com/textlint-rule/sentence-splitter/blob/40b80206aa0076d1ce8beab3bf4e01c69ed710a1/src/sentence-splitter.ts#L41-L48

azu:
This PR fixed types of textlint.
It does not change the behavior of textlint.

## Changes

- Remove `[index:string]: any;` from `TxtNode`
- Plugin's `preProcess` should return `TxtDocumentNode`
  - Previously, it allow return `TxtNode`.

For plugin developer:

You may be necessary to change the type that returns preProcess as follows

```diff
-            preProcess(_text: string, _filePath: string): TxtNode {
+            preProcess(_text: string, _filePath: string): TxtDocumentNode {
```

## Related

- fix #1294

### DIFF
--- a/packages/@textlint/ast-node-types/src/NodeType.ts
+++ b/packages/@textlint/ast-node-types/src/NodeType.ts
@@ -54,8 +54,6 @@ export interface TxtNode {
     loc: TxtNodeLocation;
     // `parent` is created by runtime
     parent?: TxtParentNode;
-
-    [index: string]: any;
 }
 
 /**
--- a/packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
+++ b/packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
@@ -1,5 +1,5 @@
 // Plugin
-import { TxtDocumentNode, TxtNode } from "@textlint/ast-node-types";
+import type { TxtDocumentNode } from "@textlint/ast-node-types";
 
 /**
  * textlint plugin option values is object or boolean.
--- a/packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
+++ b/packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
@@ -1,5 +1,5 @@
 // Plugin
-import { TxtNode } from "@textlint/ast-node-types";
+import { TxtDocumentNode, TxtNode } from "@textlint/ast-node-types";
 
 /**
  * textlint plugin option values is object or boolean.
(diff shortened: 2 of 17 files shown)


packages/@textlint/ast-node-types/src/NodeType.ts
```diff
@@ -54,8 +54,6 @@ export interface TxtNode {
     loc: TxtNodeLocation;
     // `parent` is created by runtime
     parent?: TxtParentNode;
-
-    [index: string]: any;
```
azu, line 58: Remove `[index:string]: any;` from `TxtNode`

Previoulsy, `Omit<TxtNode, "type">` return any objects. It is bad.

packages/@textlint/types/src/Plugin/TextlintPluginModule.ts
```diff
@@ -9,7 +9,7 @@ export declare type TextlintPluginOptions = {
     [index: string]: any;
 };
 
-export type TextlintPluginPreProcessResult = TxtNode | { text: string; ast: TxtNode };
+export type TextlintPluginPreProcessResult = TxtDocumentNode | { text: string; ast: TxtDocumentNode };
```
azu, line 12: Plugin's preProcessor should returns `TxtDocumentNode`

PR #1322, opened 2024-01-08
fix(config-loader): enhance config-loader results

azu:
This is breaking change, but `@textlint/config-loader` is internal use.

### `moduleName` should be package's name

Previously, `moduleName` is a file path.

```diff
            {
                "type": "RuleInPreset",
                "ruleId": "abc/c",
                "options": false,
                "filePath": "<MODULES_DIR>/textlint-rule-preset-abc/index.js",
-                "moduleName": "<MODULES_DIR>/textlint-rule-preset-abc",
+                "moduleName": "textlint-rule-preset-abc",
                "ruleKey": "c"
            }
```

### Add `type: "Plugin"`

```diff
        "plugins": [
            {
+                "type": "Plugin",
                "pluginId": "custom",
                "plugin": {},
                "filePath": "<MODULES_DIR>/textlint-plugin-custom/index.js",
                "moduleName": "textlint-plugin-custom"
            }
        ],
```

### Add `inputModuleName: string`

Previously, the input module name is not defined in results.
In some usecase, a user want to map input to result.

For this usecase, the user can use inputModuleName <-> moduleName

- https://github.com/textlint/editor/issues/87

### DIFF
--- a/packages/@textlint/config-loader/test/snapshots/preset/output.json
+++ b/packages/@textlint/config-loader/test/snapshots/preset/output.json
@@ -7,7 +7,7 @@
                 "ruleId": "abc/a",
                 "options": true,
                 "filePath": "<MODULES_DIR>/textlint-rule-preset-abc/index.js",
-                "moduleName": "<MODULES_DIR>/textlint-rule-preset-abc",
+                "moduleName": "textlint-rule-preset-abc",
                 "ruleKey": "a"
             },
             {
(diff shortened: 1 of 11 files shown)


packages/@textlint/config-loader/test/snapshots/preset/output.json
```diff
@@ -7,23 +7,26 @@
...
                 "ruleKey": "b"
             },
             {
                 "type": "RuleInPreset",
                 "ruleId": "abc/c",
                 "options": true,
                 "filePath": "<MODULES_DIR>/textlint-rule-preset-abc/index.js",
-                "moduleName": "<MODULES_DIR>/textlint-rule-preset-abc",
+                "moduleName": "textlint-rule-preset-abc",
+                "inputModuleName": "preset-abc",
```
azu, line 29: This is an example of `ruleId` and `inputModuleName`.

Issue #1123, opened 2023-03-25
`textlint-script`: support ESM outputs


azu:
Current `textlint-scripts build` output CJS.
However, [textlint v13.0.0](https://textlint.github.io/blog/2023/01/27/textlint-13.html) support native ESM rule.

`textlint-scripts` need to support ESM outputs.

## Proposal A

Add `textlint-scripts-esm` and `textlint-scripts --esm` like [ts-node](https://github.com/TypeStrong/ts-node)

- Require to output `.mjs` because `package.json` may not have `type=module`

## Proposal B

If `package.json` has `type=module`, `textlint-scripts` output ESM

- Require to run `npm pkg set "type"="module"` before using `textlint-scripts`

azu:
`textlint-scripts test` transpile js on the fly.
It requires babel loader.

- Blocker: https://github.com/babel/babel/pull/15456
- Related: [giltayar/babel-register-esm: Node.js ESM loader similar to babel-register](https://github.com/giltayar/babel-register-esm)

📝 We can run tests without loader in js mode beause babel only transpile ES2015+ syntax.
However, in ts mode, we need to get loader like babel loader or ts-node/esm

azu:
[Customization Hooks](https://nodejs.org/api/module.html#customization-hooks) is RC.

PR #1345, opened 2024-02-01
BREAKING CHANGE: Change textlint-scirpts output target

### LINKED ISSUE #1344: Update textlint-script default target
We want to update textlint-script default target.

Current is ES2015 + Module

https://github.com/textlint/textlint/blob/9925b621433cccf3e78dcf10b7c519223bd0f239/packages/textlint-scripts/configs/babel.config.js#L6-L20

However, currently, it is common to use later functions such as Async/Await after ES2015. Therefore, I think it is better to match the default target to the node.js 18, which is the support environment of the textrint.

- #1200 
- #1336

azu:
## Summary

`textlint-script builds` output ES2022-compatible code.
Previously, the output is ES2015-compatible code.

Modern browsers and Node.js LTS support ES2022.

## Details

In this PR, the target code for textlint-script output is changed to Node.js v18.14.0 or later.

This will reduce the cost of output code size and runtime.
The default target is the minimum version of node.js supported by the text link.
In other words, it is Node.js v18.14.0>=.
Runtime(browser/node.js) require to support ES2022+.

close #1344 
refs #1200


packages/textlint-scripts/configs/babel.config.js
```diff
@@ -7,10 +7,9 @@ module.exports = {
         [
             "@babel/env",
             {
-                // For async/await support
-                // https://babeljs.io/docs/en/babel-preset-env#targetsesmodules
                 targets: {
-                    esmodules: true
+                    // Same version with textlint's supported Node.js version
+                    node: "18.14.0"
                 },
```
azu, line 13: This is main change.

- ES2015 → ES2022

packages/textlint-scripts/package.json
```diff
@@ -33,6 +33,9 @@
         "url": "https://github.com/textlint/textlint/issues"
     },
     "homepage": "https://github.com/textlint/textlint/tree/master/packages/textlint-scripts",
+    "engines": {
+      "node": ">=18.19.0"
+    },
```
azu, line 38: textlint-scriptsr equire latest Node.js 18.x.
In the future, we might implement the hook processing using the `node:module` register API, so I'm choosing the latest version, 18.19.0.

- related? #1123

PR #1357, opened 2024-02-11
add TxtCommentNode to StaticPhrasingContent

### LINKED ISSUE #1356: `Content` type does not accept `Comment` type

- `@textlint/ast-node-types` 14.0.1
- typescript 5.3.3

This is a problem of TypeScript library of textlint. All files may be affected.
The actual impact depends on the implementation of the plugin.

`TxtDocumentNode` is a root node in textlint. It has some child nodes as `children` and type of it is `Content[]` since v13.0.0.
https://github.com/textlint/textlint/blob/34317df332f8b97a0c17e54c0bfc174eab38af17/packages/%40textlint/ast-node-types/src/NodeType.ts#L130-L132
https://github.com/textlint/textlint/blob/34317df332f8b97a0c17e54c0bfc174eab38af17/packages/%40textlint/ast-node-types/src/NodeType.ts#L72-L74

textlint has `Comment` type node, and it is used by some rules such as [textlint-filter-rule-comments](https://github.com/textlint/textlint-filter-rule-comments).
However, `Content` type does not contain `Comment` type node. So, I cannot create AST contains comments.

```typescript
import { ASTNodeTypes, TxtDocumentNode } from "@textlint/ast-node-types";

function createAstWithComments(): TxtDocumentNode {
    return {
        type: ASTNodeTypes.Document,
        children: [
            {
                type: ASTNodeTypes.Comment,
                value: " This is a comment",
            },
        ],
    };
}
```

Above source causes compilation error as follows:

```text
src/sample.ts:8:17 - error TS2322: Type 'ASTNodeTypes.Comment' is not assignable to type '"Code" | "CodeBlock" | "Html" | "Paragraph" | "Header" | "HorizontalRule" | "BlockQuote" | "List" | "Table" | "ListItem" | "TableRow" | "TableCell" | "Link" | "Str" | "Emphasis" | "Strong" | "Delete" | "Break" | "Image"'.

8                 type: ASTNodeTypes.Comment,
                  ~~~~

  node_modules/@textlint/ast-node-types/lib/src/NodeType.d.ts:167:5
    167     type: "Code";
            ~~~~
    The expected type comes from property 'type' which is declared here on type 'Content'
```

AST containing Comment can be created. Or, the correct way to construct an AST containing Comment is provided.

pddg:
fix #1356

### DIFF
--- a/packages/@textlint/ast-node-types/src/NodeType.ts
+++ b/packages/@textlint/ast-node-types/src/NodeType.ts
@@ -125,7 +125,8 @@ export type StaticPhrasingContent =
     | TxtHtmlNode
     | TxtCodeNode
     | TxtBreakNode
-    | TxtImageNode;
+    | TxtImageNode
+    | TxtCommentNode;
 
 export interface TxtDocumentNode extends TxtParentNode {
     type: "Document";


azu:
Thanks!

PR #1383, opened 2024-05-06
docs: deprecated `--parallel` and `--max-concurrency` in CLI

chick-p:
Related: https://github.com/textlint/textlint/pull/1338


azu:
You're right, Thanks!

PR #1415, opened 2024-08-12
Tiny readdir glob gitignore

azu:
tiny-readdir-glob-gitignore ver


azu:
closed by #1414 

Tiny readdir glob can not accept file path as glob

PR #1447, opened 2025-01-01
fix(textlint): update file-entry-cache@^10

azu:
fix deprecated warning of `file-entry-cache` package.

```
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated rimraf@2.6.3: Rimraf versions prior to v4 are no longer supported
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
```


packages/textlint/src/engine/execute-file-backers/cache-backer.ts
```diff
@@ -1,26 +1,39 @@
...
-         * @type {boolean}
-         */
+    constructor(private config: CacheBackerOptions) {
         this.isEnabled = config.cache;
-        this.fileCache = fileEntryCache.create(config.cacheLocation);
+        const filename = path.basename(config.cacheLocation);
+        const cacheDir = path.dirname(config.cacheLocation);
+        // use the metadata for cache instead of the file content
+        // TODO: if we want to reuse the cache in CI, we should use the file content cache and save relative path into the cache
+        this.fileCache = fileEntryCache.create(filename, cacheDir, false);
```
azu, line 36: It aim to preserve current behavior.

PR #1515, opened 2025-05-23
refactor: remove unused imports

noritaka1166:
remove unused imports

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -20,7 +20,10 @@
     "@docusaurus/core": "^3.7.0",
     "@docusaurus/plugin-client-redirects": "^3.7.0",
     "@docusaurus/preset-classic": "^3.7.0",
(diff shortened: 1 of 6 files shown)


website/package.json
```diff
@@ -20,10 +20,7 @@
     "@docusaurus/core": "^3.7.0",
     "@docusaurus/plugin-client-redirects": "^3.7.0",
     "@docusaurus/preset-classic": "^3.7.0",
-    "@mdx-js/react": "^3.1.0",
-    "prism-react-renderer": "^2.4.1",
-    "react": "^19.1.0",
-    "react-dom": "^19.1.0"
```
azu, line 26: Actually, docusaurus depended on react.

It seems strange to make it transitive dependency and implicit.

Issue #1498, opened 2025-05-03
textlint --mcp


azu:
ESLint support `--mcp` flag in ESLint v9.26.0.
 
- [ESLint v9.26.0 released - ESLint - Pluggable JavaScript Linter](https://eslint.org/blog/2025/05/eslint-v9.26.0-released/)

Perhaps textlint could be supported as well.

## Objective

- Implement `textlint --mcp` as an experimental feature
- Running `textlint --mcp` starts the MCP server.

From the user's point of view.

``` 
npx textlint --mcp 
```

will start MCP's server and textlint can be started as MCP server mode.

- [MCP Server Setup - ESLint - Pluggable JavaScript Linter](https://eslint.org/docs/latest/use/mcp) is a good reference!

### Architecture (Idea)

Separate MCP implementation as a package.

- Develop in the `@textlint/mcp` package.
- Start a server in the `@textlint/mcp` package when executed with the `textlint --mcp` command

This makes it easier to develop the MCP server independently.
On the other hand, users will be able to use the MCP server simply by installing `textlint`.
(A separated CLI is ideal, but in practice it is difficult for users to use (recall that @babel/cli and babel were separated in the past).)

I would like to propose to separate the packages as a compromise for this.

- https://deepwiki.com/textlint/textlint
- 📝 https://deepwiki.com/search/how-to-implement-textlint-mcp_b7e749b9-8776-4e88-a0b5-6534f9840c3a No review yet

### MCP tools

Perhaps the following tools could be provided by textlitn MCP.

- `lintFile`
- `lintText`
- `fixFile`
- `fixText`

more? 

### New Dependencies

- [modelcontextprotocol/typescript-sdk](https://github.com/modelcontextprotocol/typescript-sdk)
- [Zod](https://zod.dev/)

`typescript-sdk` required `zod`.

I think this is the downside of this proposal.

### Notes

The advantage of including this functionality in the core of textlint is that it gives access to the raw information that textlit has internally about parsing and Lint results.
I believe that by including the relevant ASTs in the tool's results, LLM can extend the context more.

For example, playwright's "Copy prompt" includes machine readable information (Aria Snapshot information is included)

- https://github.com/microsoft/playwright/releases/tag/v1.51.0

There have been cases reported that AI Aggent does not work well with just a simple textlint error result.

- [An attempt to automate entry writing by using Cline to verbalize blog style and tone - Diary of a Science Student](https://kiririmode.hatenablog.jp/entry/20250303/1740985951)

[...]

PR #1520, opened 2025-06-06
feat: add `@textlint/mcp` and add `--mcp` option to textlint CLI

chick-p:
This PR includes some improvements and verification related to the `mcp` package.  
Ref. #1498

## Verification
I confirmed the behavior with the following steps:

1. Install `textlint` locally:
   ```shell
   cd packages/@textlint/mcp/
   npm install -g .
   ```
1. Run MCP Inspector:
    ```shell
    npx -y @modelcontextprotocol/inspector npx textlint --mcp
    ```
1. Open http://127.0.0.1:6274
1. Click [Connect] and select [LintText]
1. Enter values for text and stdinFilename
1. Click [Run Tool] to execute
   

## Questions
I wasn't sure how to add a test case for invalid files in the mcp server tests.
I'd appreciate any guidance or examples on how to approach this.

### DIFF
--- a/packages/@textlint/mcp/package.json
+++ b/packages/@textlint/mcp/package.json
@@ -35,6 +35,7 @@
         "clean": "rimraf lib/ module/",
         "format": "prettier --write \"**/*.{js,jsx,ts,tsx,css}\"",
         "prepack": "npm run --if-present build",
+        "test": "mocha",
         "watch": "tsc -b --watch"
     },
     "prettier": {
--- a/packages/@textlint/mcp/package.json
+++ b/packages/@textlint/mcp/package.json
@@ -1,0 +1,61 @@
+{
+    "name": "@textlint/mcp",
+    "version": "14.7.2",
+    "description": "MCP Server for textlint",
+    "keywords": [
+        "textlint",
+        "typescript",
+        "mcp"
+    ],
+    "homepage": "https://github.com/textlint/textlint/tree/master/packages/@textlint/mcp/",
+    "bugs": {
+        "url": "https://github.com/textlint/textlint/issues"
+    },
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/textlint/textlint.git"
+    },
+    "license": "MIT",
+    "author": "azu",
+    "type": "commonjs",
+    "main": "lib/src/index.js",
+    "types": "lib/src/index.d.ts",
+    "directories": {
+        "lib": "lib",
+        "test": "test"
+    },
+    "files": [
+        "bin/",
+        "lib/",
+        "src/",
+        "!*.tsbuildinfo"
+    ],
+    "scripts": {
+        "build": "tsc -b",
+        "clean": "rimraf lib/ module/",
+        "format": "prettier --write \"**/*.{js,jsx,ts,tsx,css}\"",
+        "prepack": "npm run --if-present build",
+        "watch": "tsc -b --watch"
+    },
+    "prettier": {
+        "printWidth": 120,
+        "singleQuote": false,
+        "tabWidth": 4
+    },
+    "dependencies": {
+        "@modelcontextprotocol/sdk": "^1.12.1",
+        "textlint": "^14.7.2",
+        "zod": "^3.25.51"
+    },
+    "devDependencies": {
+        "@textlint/resolver": "^14.7.2",
+        "@types/node": "^18.19.9",
+        "prettier": "^2.7.1",
+        "rimraf": "^6.0.1",
+        "ts-node": "^10.9.2",
+        "typescript": "^5.3.3"
+    },
+    "publishConfig": {
+        "access": "public"
+    }
+}
--- a/packages/textlint/src/cli.ts
+++ b/packages/textlint/src/cli.ts
@@ -76,11 +76,26 @@ export const cli = {
             const descriptor = await loadDescriptor(currentOptions);
             Logger.log(JSON.stringify(descriptor, null, 4));
             return Promise.resolve(0);
+        } else if (currentOptions.mcp) {
+            const { server: mcpServer } = require("@textlint/mcp");
+            const { StdioServerTransport } = require("@modelcontextprotocol/sdk/server/stdio.js");
+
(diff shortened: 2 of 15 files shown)


packages/textlint/src/cli.ts
```diff
@@ -76,11 +76,26 @@ export const cli = {
             const descriptor = await loadDescriptor(currentOptions);
             Logger.log(JSON.stringify(descriptor, null, 4));
             return Promise.resolve(0);
+        } else if (currentOptions.mcp) {
+            const { server: mcpServer } = require("@textlint/mcp");
+            const { StdioServerTransport } = require("@modelcontextprotocol/sdk/server/stdio.js");
```
azu, line 81: Can you use Dynamic import instead of require?

packages/textlint/src/cli.ts
```diff
@@ -76,11 +76,26 @@ export const cli = {
...
+
+            await mcpServer.connect(new StdioServerTransport());
+            // NOTE: do not use Logger.log() because stdout use to the server transport
+            Logger.error(`Textlint MCP server is running.`);
```
azu, line 85: Question: Is it possible to use Logger.log before connecting?

```suggestion
            // NOTE: do not use Logger.log() after connect because stdout use to the server transport
            Logger.log(`Textlint MCP server is running.`);
            await mcpServer.connect(new StdioServerTransport());
```

packages/textlint/src/cli.ts
azu, line 81: What do you think of the possibility that `@textlint/mcp` itself could also provide a function to start the server?

```ts 
const { launchServer } = await import("@textlint/mcp"); 
const { closeServer } = launchServer({ /* additional options */ });

process.on("SIGINT", () => { 
  closeServer(); 
}); 
```

azu (commented):
Thanks for PR.
I've commented.

You need to add `@textlint/mcp` into https://github.com/textlint/textlint/blob/c035aad2294ce1bb78ec179c966e7174924786dc/packages/textlint/package.json#L48-L61

packages/@textlint/mcp/package.json
```diff
@@ -0,0 +1,62 @@
...
+        "singleQuote": false,
+        "tabWidth": 4
+    },
+    "dependencies": {
+        "@modelcontextprotocol/sdk": "^1.12.1",
+        "textlint": "^14.7.2",
+        "zod": "^3.25.51"
+    },
+    "devDependencies": {
+        "@textlint/resolver": "^14.7.2",
```
azu, line 52: Is this used?

azu:
> textlint@14.7.2 build
> tsc -b && tsc -b tsconfig.module.json  
>  Error: src/cli.ts(80,55): error TS2307: Cannot find module '@textlint/mcp' or its corresponding type declarations.https://github.com/textlint/textlint/actions/runs/15503006140/job/43653812792?pr=1520

This is probably caused by a circular reference in textlint <-> @textlint/mcp.
There are three possible solutions to this.

1. stop separating the packages and integrate @textlint/mcp with textlint.
2. pass an instance of `createLinter` to @textlint/mcp
3. how to use `@textlint/kernel` instead of `textlint` for @textlint/mcp

1 is simple. Just move `@textlint/mcp` to the `/mcp` directory.

2 is to create a `linter` in textlint and pass it to mcpServer.

````ts 
[const linterOptions = await makeLinterOptions(); 
const linter = createLinter(linterOptions); 
const mcpServer = await connectMcpServer({ linter} ); 
````

However, testing can be complicated.

3 is the cleanest way, but may not be beneficial in the current situation.

I feel that once it is set to 1, we should be able to move it.
3 should be able to be done later.

File: packages/textlint/src/cli.ts (hunk 2 above)
Comment (chick-p) [reply packages/textlint/src/cli.ts]: Thanks.
Even after moving the log output before the connection, running `npx @modelcontextprotocol/inspector npx textlint --mcp` still resulted in the following error:
```
Stdio transport: command=/Users/chick-p/.local/share/mise/installs/node/18.20.1/bin/npx, args=textlint,--mcp
Spawned stdio transport
Connected MCP client to backing server transport
Created client/server transports
Set up MCP proxy
Received message for sessionId 55d77644-d73a-4d64-9187-89924c7ebd0f
Error from MCP server: SyntaxError: Unexpected token T in JSON at position 0
    at JSON.parse (<anonymous>)
    at deserializeMessage (file:///Users/chick-p/.npm/_npx/5a9d879542beca3a/node_modules/@modelcontextprotocol/sdk/dist/esm/shared/stdio.js:26:44)
    at ReadBuffer.readMessage (file:///Users/chick-p/.npm/_npx/5a9d879542beca3a/node_modules/@modelcontextprotocol/sdk/dist/esm/shared/stdio.js:19:16)
    at StdioClientTransport.processReadBuffer (file:///Users/chick-p/.npm/_npx/5a9d879542beca3a/node_modules/@modelcontextprotocol/sdk/dist/esm/client/stdio.js:127:50)
    at Socket.<anonymous> (file:///Users/chick-p/.npm/_npx/5a9d879542beca3a/node_modules/@modelcontextprotocol/sdk/dist/esm/client/stdio.js:98:22)
    at Socket.emit (node:events:517:28)
    at addChunk (node:internal/streams/readable:368:12)
    at readableAddChunk (node:internal/streams/readable:341:9)
    at Readable.push (node:internal/streams/readable:278:10)
    at Pipe.onStreamRead (node:internal/stream_base_commons:190:23)
Received message for sessionId 55d77644-d73a-4d64-9187-89924c7ebd0f
Received message for sessionId 55d77644-d73a-4d64-9187-89924c7ebd0f
Received message for sessionId 55d77644-d73a-4d64-9187-89924c7ebd0f
```

In https://github.com/textlint/textlint/pull/1522, I tried removing the log output entirely. What do you think about this approach?

chick-p:
@azu Thank you for your advice.

Regarding the circular dependency, I tried addressing it using the first approach you suggested—moving it to `/mcp`.
It seemed like a simple and clean solution.
Since the diff turned out to be quite large, I’ve recreated the changes in a new PR:
https://github.com/textlint/textlint/pull/1522

PR #1540, opened 2025-06-14
Use pnpm catalog for dependencies

azu:
Switch all dependencies and devDependencies to use the pnpm catalog format for better package management.

- https://pnpm.io/catalogs


azu:
https://github.com/lerna/lerna/issues/4055
lerna does not support catalog

ghiscoding:
> [lerna/lerna#4055](https://github.com/lerna/lerna/issues/4055) lerna does not support catalog

It's supported in Lerna-Lite since a few months already and I use it in a few of my projects 😉

PR #1715, opened 2025-08-30
CI: fix release workflow

azu:


.github/workflows/release.yml
```diff
@@ -112,7 +112,7 @@ jobs:
         with:
           tag_name: v${{ env.PACKAGE_VERSION }}
           # Copy Pull Request's tile and body to Release Note
-          release_name: ${{ github.event.pull_request.title }}
+          name: ${{ github.event.pull_request.title }}
```
azu, line 115: name is correct

PR #1722, opened 2025-09-05
docs: add info severity level to configuring page

3w36zj6:
I added a description of the info severity level, which was introduced in [v15.1.0](https://github.com/textlint/textlint/releases/tag/v15.1.0), to the documentation's configuring page.


azu (approved):
nice catch

Issue #1122, opened 2023-03-24
Suggest API


azu:
- https://github.com/textlint/textlint/discussions/833

## Purpose

We want to support suggestion on editor like VSCode.
`--fix` is useful but it provide only a single suggestion.
Also, User can not excludes some fix that is not want to fix.

Some rules want to provide multiple suggestions.
This proposal will near to current `fix` property.

LLM (Large Language Model) checking is becoming more and more practical.
On the other hand, machine learning-based is not an automatic modification like Lint, but is more appropriate for users to select modifications.　This is because it is a matter of probability.

This requires user-oriented interactivity.

## Proposal

- Add `suggest` property to RuleError
    - `suggest` should have an array of object that has `id` and `fix` and `message` property
- The `suggest` property will be into `TextlintMessage`
- `textlint` cli may does not provide fixing function the suggest at first
    - This fixing need to use `id` of suggestion.(ESLint does not have the `id`)
    - [x] It will need new funciton in kernel to fix text with message
        - https://github.com/textlint/textlint/tree/master/packages/%40textlint/source-code-fixer has the function
        - It is already separated module

```js
const reporter = (context) => {
    const { Syntax, RuleError, fixer, report, getSource } = context;
    return {
        [Syntax.Str](node) {
            const text = getSource(node);
            // "You fix this"
            //      ^^^
            const matchRegexp = /\bfix\b/;
            if (!matchRegexp.test(text)) {
                return;
            }
            // found "fixable" error
            const index = text.search(matchRegexp);
            const length = "fix".length;
            report(
                node,
                new RuleError("Replaced", {
                    index,
                    suggest: [
                        {
                            id: "pattern1",
                            message: "pattern 1",
                            fix: fixer.replaceTextRange([index, index + length], "fixed 1")
                        },
                        {
                            id: "pattern2",
                            message: "pattern 2",
                            fix: fixer.replaceTextRange([index, index + length], "fixed 2")
                        }
                    ]
                })
            );
        }
    };
};
export default {
    linter: reporter,
    fixer: reporter
};
```

TODO:

- [ ] Suggest API is required?
  - `new RuleSuggest(message, { suggest: [ ... ] })` or `new RuleError(message, { serverity: "info" })`
  - Current `RuleError` will be an error

Similar Works:

- ESLint
    - https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions

azu:
[...]

PR #1801, opened 2025-10-25
feat(kernel): add suggestions API

3w36zj6:
cf. https://github.com/textlint/textlint/issues/1122

This pull request adds support for reporting and handling suggestions in linting results, enabling rules to provide multiple suggested fixes for a single problem. The implementation includes changes to core logic, error handling, and corresponding tests to ensure suggestions are properly processed and output.

These changes introduce a flexible and extensible way for rules to offer multiple, context-aware suggestions to users, improving the overall usability and power of the linting engine.

### DIFF
--- a/packages/@textlint/kernel/test/helper/SuggestionRule.ts
+++ b/packages/@textlint/kernel/test/helper/SuggestionRule.ts
@@ -2,7 +2,7 @@ import { TxtNode } from "@textlint/ast-node-types";
 import type { TextlintRuleContext, TextlintRuleModule, TextlintRuleReporter } from "@textlint/types";
 
 export const report: TextlintRuleReporter = (context: Readonly<TextlintRuleContext>) => {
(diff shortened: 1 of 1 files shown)


packages/@textlint/kernel/test/helper/SuggestionRule.ts
```diff
@@ -0,0 +1,40 @@
...
+                new RuleError("found suggestion", {
+                    suggestions: [
+                        {
+                            id: "suggestion-1",
+                            message: "use alternative",
+                            fix: {
+                                isAbsolute: false,
+                                range: [0, 1],
+                                text: "X"
+                            }
```
azu: Can use you `fixer` instead of raw object?

azu (commented):
Sorry to delay.
Almost LGTM, but I commented.

azu (approved):

PR #2038, opened 2026-05-13
feat(ast-node-types): make TxtNode properties readonly

### LINKED ISSUE #489: textlint-tester: validate AST modification
textlint rule should not modify AST.
But, this is implicitly defined.

We want to validate AST modification in the textlint-tester

ESLint does this test in ESLint's rule tester.

https://github.com/eslint/eslint/blob/33ca1ea67e41a05ff1283f6be36553db3904ec1f/lib/testers/rule-tester.js#L398-L399

## Tasks

- [ ] TBD

## Related

- Disscuss textlint-tester extensible #488

3w36zj6:
close #489 

This PR aims to make the properties of TxtNode in ast-node-types readonly, encouraging rule authors to avoid unintended AST mutation.

This approach looks useful for rule authors, but I'm concerned that it may make type-safe AST manipulation more cumbersome for plugin authors. It may be worth considering alternatives, such as using a ReadonlyTxtNode type specifically for nodes passed to rule visitors.

### DIFF
--- a/docs/rule.md
+++ b/docs/rule.md
@@ -10,6 +10,9 @@ textlint's AST(Abstract Syntax Tree) is defined at the page.
 
 Each rule are represented by an object with some properties.
 The properties are equivalent to AST node types from TxtNode.
+Rule visitor receives readonly AST nodes in TypeScript.
+Rules should not mutate nodes directly, such as assigning to `node.value` or pushing to `node.children`.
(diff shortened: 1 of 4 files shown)


azu (approved):
LGTM


In textlint's policy, type-only changes can be treated as minor.
https://github.com/textlint/textlint#semantic-versioning-policy

PR #2078, opened 2026-06-20
refactor(ast-node-types): simplify TypeofTxtNode with a lookup map

314systems:
Replace the long nested conditional type with a TxtNodeTypeMap interface
and an indexed access lookup. Behavior is unchanged for both single and
union node types, with AnyTxtNode kept as the fallback.

Also fix doc comment typos (TxtNodeTYpe, TxtNodeTypeOfNode).


azu (approved):
LGTM

PR #2082, opened 2026-07-01
fix: prevent duplicated register in textlint-scripts

### LINKED ISSUE #781: textlitn-scripts:  Prevent duplicated register
From https://github.com/textlint/textlint-scripts/issues/38

```
--require textlint-scripts/register-ts
--require textlint-scripts/register-ts
```

throw error TS7006: Parameter 'obj' implicitly has an 'any' type.

mmustafasenoglu:
## Description

This PR fixes duplicate register errors that occur when `--require textlint-scripts/register-ts` or `--require textlint-scripts/register` is specified multiple times.

### Problem

When a user has `--require textlint-scripts/register-ts` in both their `.mocharc` file and their `package.json` scripts, the register gets called twice, causing errors like:

```
error TS7006: Parameter 'obj' implicitly has an 'any' type.
```

This happens because the register files don't check if they've already been loaded.

### Fix

Added global variable guards to prevent duplicate registration:

1. **`register-ts.js`**: Added `global.__textlint_scripts_register_ts_loaded` check
2. **`register.js`**: Added `global.__textlint_scripts_register_loaded` check

### Changes

- **`packages/textlint-scripts/register-ts.js`**: Added duplicate registration guard
- **`packages/textlint-scripts/register.js`**: Added duplicate registration guard

### Testing

- Verified that running with duplicate `--require` no longer throws errors
- Existing functionality remains unchanged

Fixes #781

### DIFF
--- a/packages/textlint-scripts/register-ts.js
+++ b/packages/textlint-scripts/register-ts.js
@@ -1,2 +1,6 @@
 // --require textlint-scripts/register-ts
-require("./configs/ts-node-register");
+// Prevent duplicate registration
+if (!global.__textlint_scripts_register_ts_loaded) {
+    global.__textlint_scripts_register_ts_loaded = true;
+    require("./configs/ts-node-register");
+}
--- a/packages/textlint-scripts/register-ts.js
+++ b/packages/textlint-scripts/register-ts.js
@@ -1,6 +1,9 @@
 // --require textlint-scripts/register-ts
-// Prevent duplicate registration
+// Prevent duplicate registration.
+// Node.js require cache handles most cases, but in monorepos with symlinked
+// node_modules or npm link, the same logical module may resolve to different
+// absolute paths, bypassing the cache. This global flag covers that edge case.
 if (!global.__textlint_scripts_register_ts_loaded) {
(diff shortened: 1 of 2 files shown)


mmustafasenoglu:
@devin-ai-integration Great catches! Fixed both issues:

**Bug fix**: Moved guard flag assignment **after** `require()` so if registration throws (missing ts-node/@babel/register), subsequent retries can still attempt registration.

**Documentation**: Added explanation of why the global flag is needed on top of Node.js require cache - covers monorepo/symlink scenarios where the same module resolves to different absolute paths.

Thanks for the thorough review!

mmustafasenoglu:
@devin-ai-integration Thanks for the thorough review! Both bugs you identified have been addressed:

**Fixed:** Guard flag is now set AFTER `require()` succeeds in both `register-ts.js` and `register.js`, so failed registrations can be retried.

**Re: require cache:** You're right that Node.js cache handles most cases. Added a comment explaining the global flag is needed for monorepo/symlink scenarios where the same module resolves to different absolute paths.

All issues marked as resolved. Thanks for catching these!

Issue #1307, opened 2024-01-02
Meta: ECMAScript Modules


azu:
This is Meta Issue that overview for ESM.

## Core

- Current(v13): textlint is CJS
- Future: textlint will be ESM

## ✅ Rules/Plugins

- Current(v13): textlint can load CJS and ESM rules/plugins/formatter
  - [Release v13.0.0 · textlint/textlint](https://github.com/textlint/textlint/releases/tag/v13.0.0)

## textlint-scripts

- Current(v13): textlint-scripts output CJS
- Future: textlint-scripts output ESM
- #1123 

## editor

- Current: textlint-editor uses Web Worker code
- Future: textlint-editor will use ESM rules/plugins directly or use Web Worker type=module
- Related: https://github.com/textlint/editor/issues/84

PR #2109, opened 2026-08-09
feat: migrate built-in plugins to ESM and upgrade remark dependencies

3w36zj6:
Related to #1307.

## Changes

The remark dependencies used internally by the Markdown plugin have become outdated, preventing the parser from keeping pace with modern Markdown specifications. This would also make it difficult to support well-known extensions such as CJK-friendly syntax, because many modern remark plugins are ESM-only.

To address these limitations, this PR migrates textlint's built-in Markdown and plain-text plugins to ESM and upgrades the remark dependencies used by `@textlint/markdown-to-ast`.

### textlint-plugin-markdown and markdown-to-ast

This change upgrades `unified`, `remark-parse`, `remark-gfm`, and `remark-frontmatter` to their ESM-only major versions. To preserve the existing AST location behavior, it disables the GFM autolink literal transforms after the GFM extension is registered. The previous implementation disabled these transforms by overriding `mdast-util-gfm-autolink-literal/from-markdown` before loading `remark-gfm`, but that subpath is no longer exported by the upgraded ESM package.

The parser upgrade changes some AST output to follow the newer CommonMark and GFM implementations.

### textlint-plugin-text and text-to-ast

Although this migration is otherwise unrelated to the remark upgrade, `@textlint/textlint-plugin-text` is also migrated to ESM so that consumers can handle both built-in plugins consistently, as described below.

### textlint core

The `textlint` and `@textlint/legacy-textlint-core` packages remain CommonJS. They now load the ESM built-in plugins with dynamic `import()` before constructing the kernel options. The legacy core captures the current descriptor before loading the plugins so that later configuration changes do not affect an in-progress lint or fix operation.

### textlint-tester

`textlint-tester` also remains CommonJS and keeps its test registration API synchronous. It loads the built-in plugins dynamically and waits for the completed descriptor when each lint or fix operation runs. The built-in descriptor is prepended to the test descriptor to preserve the existing plugin order.

## Notes

The built-in plugin packages are now ESM-only. The CommonJS `textlint`, `textlint-tester`, and `@textlint/legacy-textlint-core` packages remain compatible by loading these plugins with dynamic `import()`. CommonJS code that loads either built-in plugin package directly must use dynamic `import()` instead of `require()`.


packages/@textlint/markdown-to-ast/test/fixtures/list-continuation.text/output.json
```diff
@@ -118,46 +118,64 @@
...
+                                    "range": [
+                                        16,
+                                        31
+                                    ],
+                                    "raw": "[foo][]\n[foo]: "
+                                },
+                                {
+                                    "type": "Link",
+                                    "title": null,
+                                    "url": "http://google.com",
```
azu, line 142: It seems like LinkReference support has been discontinued, right?
Well, from a textlint perspective, it's not necessarily a bad thing since it reduces the number of custom handling rules, but since symbols like `[` are now included in `value`, some rules might break. (Though I doubt it will happen very often.)

packages/@textlint/markdown-to-ast/test/fixtures/lists-with-code-and-rules.text/output.json
```diff
@@ -110,22 +110,22 @@
...
+                                                                    "column": 18
                                                                 }
                                                             },
                                                             "range": [
                                                                 25,
-                                                                28
+                                                                100
                                                             ],
-                                                            "raw": "one"
+                                                            "raw": "one\n        - two\n            - three\n            - four\n            - five"
```
azu, line 128: 📝 This looks like a bug fix, so it should match GFM's rendering as well.

azu (approved):
The AST results have changed, and I'm a little concerned about LinkReference; I wonder if this is an intentional change to the remark.
I'm a bit unsure whether we should maintain it with some option or just merge it as is.

Personally, I think it would be okay to merge this change as is.
@3w36zj6 Do you have any opinions?

packages/@textlint/markdown-to-ast/test/fixtures/list-continuation.text/output.json
3w36zj6: LinkReference support itself has not been discontinued. The current parser still emits LinkReference when a valid reference definition exists, and the existing fixtures continue to cover [those cases](https://github.com/3w36zj6/textlint/blob/710a09e717fc131abe82ced8cf8e3f1ce0523770/packages/%40textlint/markdown-to-ast/test/fixtures/links-reference-style.text/input.md).

This particular input changed because `[foo]: ...` immediately follows an open paragraph in the list item without a blank line. According to [Example 182](https://github.github.com/gfm/#example-182) in the [GitHub Flavored Markdown Spec](https://github.github.com/gfm/), a link reference definition cannot interrupt a paragraph. Therefore, `[foo]: ...` is not recognized as a `Definition`, and `[foo][]`, which consequently has no matching definition, is treated as plain text. I believe this behavior is correct. Adding a blank line before the definition still produces LinkReference and Definition nodes.

I think we should keep the new AST because it matches current CommonMark/GFM, and document this as a breaking parser change rather than adding a compatibility post-process.

PR #2113, opened 2026-08-22
feat(textlint-plugin-markdown): add CJK-friendly emphasis and strikethrough support

3w36zj6:
Related to https://github.com/textlint-ja/textlint-rule-preset-ja-spacing/issues/132.

This PR adds an opt-in `cjkFriendly` option to `@textlint/textlint-plugin-markdown` and `@textlint/markdown-to-ast`.

CommonMark's delimiter rules can prevent emphasis from being recognized in Chinese, Japanese, and Korean text when punctuation appears immediately inside the closing delimiter and a non-space character follows it. The same delimiter behavior also applies to GFM strikethrough. This option allows textlint to parse these cases using the [`remark-cjk-friendly`](https://github.com/tats-u/markdown-cjk-friendly) plugins. CJK-friendly emphasis is supported by several widely used Markdown implementations and documentation frameworks, including GitLab Flavored Markdown, Rspress, and VitePress.

The option is disabled by default, so existing Markdown parsing behavior is preserved unless `cjkFriendly` is explicitly set to `true`.

### DIFF
--- a/packages/@textlint/textlint-plugin-markdown/README.md
+++ b/packages/@textlint/textlint-plugin-markdown/README.md
@@ -4,46 +4,71 @@ Markdown support for [textlint](https://github.com/textlint/textlint "textlint")
 
 ## Installation
 
-    npm install @textlint/textlint-plugin-markdown
+```sh
+npm install @textlint/textlint-plugin-markdown
+```
 
 ## Usage
 
 Built-in support on textlint.
 No need configuration.
 
-------
+---
 
 Following config is set by default.
 
-```
+```json
 {
-    "plugins": {
-        "@textlint/markdown": true
-    }
+  "plugins": {
+    "@textlint/markdown": true
+  }
 }
 ```
 
 ## Options
 
-- `extensions`: `string[]`
-    - Additional file extensions for markdown
-    
-For example, if you want to treat [MDX](https://github.com/mdx-js/mdx) as markdown, put following config to `.textlintrc`    
+- `extensions`: `string[]` (default: `[]`)
+  - Additional file extensions for markdown.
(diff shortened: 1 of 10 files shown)


packages/@textlint/textlint-plugin-markdown/README.md
```diff
@@ -4,46 +4,71 @@ Markdown support for [textlint](https://github.com/textlint/textlint "textlint")
...
-        }
+  "plugins": {
+    "@textlint/markdown": {
+      "extensions": [".mdx"]
     }
+  }
 }
 ```
 
+For example, the following Markdown is parsed as emphasis and strikethrough when `cjkFriendly` is enabled:
```
azu: Does this README help users understand when they should use this option?
I found it a little difficult to understand.

`extensions: [".mdx"]` is something you enable when you want to use MDX,
It would be good to include a README explaining when this option should be set to `true`. (The shorter and simpler the explanation, the better.)

The current README only describes the features and doesn't explain the use cases.

azu:
@codex review

packages/@textlint/textlint-plugin-markdown/README.md
3w36zj6: Thanks for the feedback.

I updated the README to explain that `cjkFriendly` should be enabled when the target Markdown renderer supports CJK-friendly parsing rules for emphasis and GFM strikethrough. I also added GitLab Flavored Markdown, Rspress, and VitePress as concrete examples of implementations that support these rules. In addition, I clarified the default CommonMark/GFM behavior and simplified the examples for Japanese, Chinese, and Korean.

